PR #17, opened 2016-03-24
global: addition of community creation signal

krzysztof:
* Adds community creation signal and integrates with OAI-PMH sets.

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -38,15 +38,16 @@ COMMUNITIES_SORTING_OPTIONS = [
 """Possible communities sorting options."""
 
 COMMUNITIES_DEFAULT_SORTING_OPTION = 'ranking'
+"""Default sorting option."""
 
-COMMUNITIES_USE_OAI = False
+COMMUNITIES_OAI_FORMAT = 'user-{community_id}'
+"""String template for the community OAISet 'spec'."""
+
+COMMUNITIES_OAI_ENABLED = False
 """Using OAIServer if available."""
-
-COMMUNITIES_OAI_PREFIX = 'user-'
-"""OAISet 'spec' prefix for community."""
 
 try:
     pkg_resources.get_distribution('invenio_oaiserver')
-    COMMUNITIES_USE_OAI = True
+    COMMUNITIES_OAI_ENABLED = True
 except pkg_resources.DistributionNotFound:  # pragma: no cover
     pass
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -21,6 +21,8 @@
 
 from __future__ import unicode_literals
 
+import pkg_resources
+
 COMMUNITIES_REQUEST_EXPIRY_TIME = 7  # Default expiry time delta in days
 
 COMMUNITIES_LOGO_EXTENSIONS = ['.png', '.jpg', '.jpeg']
@@ -36,3 +38,15 @@ COMMUNITIES_SORTING_OPTIONS = [
 """Possible communities sorting options."""
 
 COMMUNITIES_DEFAULT_SORTING_OPTION = 'ranking'
+
+COMMUNITIES_USE_OAI = False
+"""Using OAIServer if available."""
+
+COMMUNITIES_OAI_PREFIX = 'user-'
+"""OAISet 'spec' prefix for community."""
+
+try:
+    pkg_resources.get_distribution('invenio_oaiserver')
+    COMMUNITIES_USE_OAI = True
+except pkg_resources.DistributionNotFound:  # pragma: no cover
+    pass
--- a/invenio_communities/models.py
+++ b/invenio_communities/models.py
@@ -40,7 +40,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types import UUIDType
 
-from invenio_communities.errors import InclusionRequestExistsError, \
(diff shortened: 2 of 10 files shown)


krzysztof:
- [ ] Before/After community created signal
- [ ] Integration with OAI sets as a configurable
- [ ] Receiver for OAI sets

krzysztof:
Should this be configurable with config variable or with some check for ``invenio-oaiserver`` app ? (Probably in a lazy way)

invenio_communities/receivers.py
```diff
@@ -27,3 +29,14 @@ def inject_provisional_community(sender, json=None, record=None):
...
+
+
+def create_oaipmh_set(mapper, connection, community):
+    """Signal for creating OAI-PMH sets during community creation."""
+    from invenio_oaiserver.models import OAISet
+    with db.session.begin_nested():
+        obj = OAISet(spec=community.id,
+                     name=community.title,
+                     description=community.description,
+                     search_pattern="community:{0}".format(community.id))
```
jirikuncar: `search_pattern='community:"{0}"'.format(community.id))` (note the double quotes)

tests/test_invenio_communities.py
```diff
@@ -145,3 +146,22 @@ def test_model_featured_community(app):
...
+    """Test the OAI-PMH Sets creation."""
+    with app.app_context():
+        # Create a user
+        user1 = create_test_user()
+        comm1 = Community(id='comm1',
+                          id_user=user1.id,
+                          title='Title1',
+                          description='Description1')
+        db.session.add(comm1)
+        assert OAISet.query.count() == 0  # TODO: Strange SQLA issue
```
jirikuncar: Probably because you haven't commit the DB session.

tests/test_invenio_communities.py
```diff
@@ -145,3 +146,22 @@ def test_model_featured_community(app):
...
+    """Test the OAI-PMH Sets creation."""
+    with app.app_context():
+        # Create a user
+        user1 = create_test_user()
+        comm1 = Community(id='comm1',
+                          id_user=user1.id,
+                          title='Title1',
+                          description='Description1')
+        db.session.add(comm1)
+        db.session.commit()
```
krzysztof: @jirikuncar This works but is not entirely intuitive.

jirikuncar: @krzysztof there is a study material for you http://docs.sqlalchemy.org/en/rel_0_9/orm/session_basics.html.

> the `Session` establishes all conversations with the database and represents a “holding zone” for all the objects which you’ve loaded or associated with it during its lifespan

krzysztof: Thank you for the informative reference but that wasn't where my confusion came from ;-)

I wrongly assumed that "after_insert" signal is sent after the ``session.add``, rather than SQL INSERT, thus I expected receiver function to be called without the need for session to commit.

krzysztof:
@jirikuncar Added community removal signal.

invenio_communities/models.py
```diff
@@ -320,6 +320,17 @@ def filter_communities(cls, p, so):
             query = query.order_by(db.desc(cls.ranking))
         return query
 
+    @property
+    def oaiset_spec(self):
```
jirikuncar: IMHO this should be an utility function.

jirikuncar: you already have it - `get_oaiset_spec`

invenio_communities/receivers.py
```diff
@@ -17,13 +17,43 @@
...
     q = InclusionRequest.query.filter_by(id_record=record.id)
     provisional_communities_ids = [r.id_community for r in q]
     json['provisional_communities'] = provisional_communities_ids
+
+
+def create_oaipmh_set(mapper, connection, community):
+    """Signal for creating OAI-PMH sets during community creation."""
+    from invenio_oaiserver.models import OAISet
+    with db.session.begin_nested():
+        obj = OAISet(spec=community.oaiset_spec,
```
jirikuncar: `get_oaiset_spec`

invenio_communities/receivers.py
```diff
@@ -17,13 +17,43 @@
...
+                     search_pattern='community:"{0}"'.format(community.id))
+        db.session.add(obj)
+
+
+def destroy_oaipmh_set(mapper, connection, community):
+    """Signal for creating OAI-PMH sets during community creation."""
+    from invenio_oaiserver.models import OAISet
+    with db.session.begin_nested():
+        oaiset = OAISet.query.filter_by(
+            spec=community.oaiset_spec).one_or_none()
```
jirikuncar: get_oaiset_spec

invenio_communities/models.py
krzysztof: Actually that was a leftover, which was supposed to be removed on behalf of the property. But I can remove the property instead.

invenio_communities/config.py
```diff
@@ -36,3 +38,18 @@
 """Possible communities sorting options."""
 
 COMMUNITIES_DEFAULT_SORTING_OPTION = 'ranking'
+
+COMMUNITIES_OAI_NAMESPACE = 'user'
+"""OAISet 'spec' namespace for community."""
+
+COMMUNITIES_OAI_FORMAT = '{namespace}-{community_id}'
```
jirikuncar: IMHO `COMMUNITIES_OAI_FORMAT = 'user-{community_id}`' would do the job (we don't need `COMMUNITIES_OAI_NAMESPACE`)

PR #21, opened 2016-04-05
REST: addition of communities information rest

JavierDelgadoFernandez:
* Addition of REST methods to retrieve communities information.

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>

### DIFF
--- a/invenio_communities/ext.py
+++ b/invenio_communities/ext.py
@@ -37,7 +37,7 @@ from .permissions import permission_factory
 from .receivers import create_oaipmh_set, destroy_oaipmh_set, \
     inject_provisional_community, new_request
 from .signals import inclusion_request_created
-from .views import blueprint
+from .views import ui_blueprint
 
 
 class InvenioCommunities(object):
@@ -52,7 +52,7 @@ class InvenioCommunities(object):
         """Flask application initialization."""
         self.init_config(app)
         app.cli.add_command(cmd)
-        app.register_blueprint(blueprint)
+        app.register_blueprint(ui_blueprint)
         app.extensions['invenio-communities'] = self
         # Register the jinja do extension
         app.jinja_env.add_extension('jinja2.ext.do')
--- a/invenio_communities/serializers/schemas/community.py
+++ b/invenio_communities/serializers/schemas/community.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2013, 2014, 2015, 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Community JSON schema."""
+
+from __future__ import absolute_import, print_function
+
+from marshmallow import Schema, fields
+
+
+class CommunitySchemaV1(Schema):
+    """Schema for a community."""
+
+    id = fields.String()
+    description = fields.String()
+    title = fields.String()
+    page = fields.String()
+    curation_policy = fields.String()
+    last_record_accepted = fields.String()
--- a/invenio_communities/views/api.py
+++ b/invenio_communities/views/api.py
@@ -1,0 +1,185 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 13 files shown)


JavierDelgadoFernandez:
### Working on:
- [ ] Proper mapping of the URL

JavierDelgadoFernandez:
@lnielsen Should it be configurable the fields that are included in the response?

JavierDelgadoFernandez:
@lnielsen The extension is not split into two applications because the API application was totally empty.

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+                        "id": "comm1"
+                    },
+                    {
+                        "id": "comm2"
+                    }
+                ]
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
+        """
+        communities = Community.query.all()
```
lnielsen:
```python
Communities.filter_communities(...)
```
this will take care of filtering deleted communities out and then we can support querying and sorting as well:

``/communities/:id?q=...&sort=...``

- ``q`` maps to ``p`` in ``filter_communities()``
- ``sort`` maps to ``so`` in ``filter_communities()``

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+        result = []
+        for community in communities:
+            result.append({
+                'id': community.id
+            })
+        return {'json': result}
```
lnielsen: I think this is likely more an issue in``ContentNegotiatedMethodView``, but it would be better to return just ``result``. You might need to call ``self.make_response`` explicitly here.

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+            }
+        abort(404)
+
+
+serializers = {'application/json': json_serializer}
+
+
+communities_view = CommunitiesResource.as_view(
+    'communities_api',
+    serializers=serializers
```
lnielsen: I think ``default_media_type`` should also be set in order to make sure we return ``application/json`` in case no ``Accept`` header was provided. Ditto for below.

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+                    "id": "comm1",
+                    "page": "",
+                    "curation_policy": ""
+                }
+
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
+            :statuscode 404: page not found
+        """
+        community = Community.query.get(community_id)
```
lnielsen: ``Communities.get()`` - this will filter out deleted communities.

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+        if community:
```
lnielsen:
```
if not community:
   abort(404)
...
```

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+                        "id": "comm2"
+                    }
+                ]
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
+        """
+        communities = Community.query.all()
+        result = []
+        for community in communities:
+            result.append({
```
lnielsen: Better just return a list of communities, then the serializer can decide how much it wants to return.

invenio_communities/views/api.py
```diff
@@ -0,0 +1,177 @@
...
+
+from invenio_communities.models import Community
+from invenio_communities.serializer import json_serializer
+
+
+blueprint = Blueprint(
+    'invenio_communities_api',
+    __name__,
+    url_prefix='/communities',
+    template_folder='templates',
```
lnielsen: I don't think ``template_folder`` and ``static_folder`` are needed.

invenio_communities/serializer.py
```diff
@@ -0,0 +1,34 @@
...
+"""REST API serializers."""
+
+import json
+
+from flask import Response
+
+
+def json_serializer(data):
+    """Return a flask response serializing the data to JSON."""
+    return Response(json.dumps(data['json']), mimetype='application/json')
```
lnielsen: Can you use Marshmallow to serialise the community instead.

invenio_communities/views/api.py
lnielsen: Note, the ``q`` and ``sort`` parameters can be extracted using ``webargs`` package. See https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/views.py#L209

lnielsen:
@JavierDelgadoFernandez Sorry for taking so long time to review. I've gone through it now. There was quite some changes to communities in the mean time. Do you mind just rebasing and fixing above issue.

invenio_communities/ext.py
```diff
@@ -29,12 +29,12 @@
 from invenio_indexer.signals import before_record_index
 from sqlalchemy.event import listen
 
+from invenio_communities.views import api_blueprint, communities_blueprint
```
jirikuncar: `from .views import ...`

invenio_communities/views/api.py
```diff
@@ -0,0 +1,193 @@
...
+                        "id": "comm1"
+                    },
+                    {
+                        "id": "comm2"
+                    }
+                ]
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
+        """
+        communities = Community.filter_communities(query, sort)
```
lnielsen: ``communities.paginate(page,per_page).items``

invenio_communities/views/api.py
```diff
@@ -0,0 +1,193 @@
...
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
+            :statuscode 404: page not found
+        """
+        community = Community.get(community_id)
+        if not community:
+            abort(404)
+        return self.make_response({
+
+            'id': community.id,
```
lnielsen: marshmallow

invenio_communities/views/api.py
```diff
@@ -0,0 +1,193 @@
...
+            missing=None,
+        ),
+        sort=fields.String(
+            location='query',
+            missing=None,
+        ),
+        page=fields.Int(
+            location='query',
+            missing=1,
+        ),
```
lnielsen: Add ``size`` (i.e. per_page)

invenio_communities/serializers/schemas/community.py
```diff
@@ -0,0 +1,40 @@
...
+from __future__ import absolute_import, print_function
+
+from marshmallow import Schema, fields
+
+
+class CommunitySchemaV1(Schema):
+    """Schema for a community."""
+
+    id = fields.String()
+    id_user = fields.Integer()
```
lnielsen: Remove

lnielsen: + add ``last_record_accepted``

Issue #23, opened 2016-04-07
emails: fix email sending


lnielsen:
Broken links

krzysztof:
Can you give more details about this?

lnielsen:
This email is not being sent: https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/invenio_communities/new_upload_email.html
Also ``{{config.CFG_SITE_URL}}`` does not exists anymore.

krzysztof:
Thanks :+1:

Issue #26, opened 2016-04-07
views: deleting community does not remove identifier from records


lnielsen:

slint:
Closing because of module refactoring

PR #29, opened 2016-04-12
global: community deletion and permissions

krzysztof:
* Adds deleting for communities.

* Adds simple permissions for owners and admin.

* Adds inclusion request mail notifications.


krzysztof:
Closes #23

Partially addresses #26 although the task for actual removal of communities in record metadata needs to be implemented

invenio_communities/config.py
```diff
@@ -23,7 +23,12 @@
 
 import pkg_resources
 
-COMMUNITIES_REQUEST_EXPIRY_TIME = 7  # Default expiry time delta in days
+COMMUNITIES_REQUEST_EXPIRY_TIME = 7
+"""Time after which the inclusion requests automatically expire (in days)."""
+
+COMMUNITIES_DELETE_HOLDOUT_TIME = 7
+"""Time after which the communities marked for deletion are """
+"""hard-deleted from the system (in days)."""
```
jirikuncar: `"""` works on multiple lines, hence starting new line with `"""` is not necessary.

invenio_communities/config.py
```diff
@@ -23,7 +23,12 @@
 
 import pkg_resources
 
-COMMUNITIES_REQUEST_EXPIRY_TIME = 7  # Default expiry time delta in days
+COMMUNITIES_REQUEST_EXPIRY_TIME = 7
```
jirikuncar: For consistency with other modules I would propose to use seconds to defining time.

invenio_communities/cli.py
```diff
@@ -86,6 +87,7 @@ def request(community_id, record_id, accept):
         db.session.commit()
     else:
         InclusionRequest.create(community=c, record=record)
+        RecordIndexer().index_by_id(record.id)
```
lnielsen: After commit

invenio_communities/config.py
krzysztof: Using timedelta instead.

invenio_communities/models.py
```diff
@@ -228,6 +261,12 @@ def add_record(self, record):
             record[communities_key] = record[communities_key] + [self.id, ]
         else:
             record[communities_key] = [self.id, ]
+        if current_app.config["COMMUNITIES_OAI_ENABLED"] and "_oai" in record:
+            oai_spec = get_oaiset_spec(self.id)
+            if "sets" in record["_oai"]:
+                record["_oai"]["sets"] = record["_oai"]["sets"] + [oai_spec, ]
```
krzysztof: Can this be added like this?

invenio_communities/models.py
```diff
@@ -228,6 +261,12 @@ def add_record(self, record):
             record[communities_key] = record[communities_key] + [self.id, ]
         else:
             record[communities_key] = [self.id, ]
+        if current_app.config["COMMUNITIES_OAI_ENABLED"] and "_oai" in record:
+            oai_spec = get_oaiset_spec(self.id)
```
jirikuncar: inveniosoftware/invenio-oaiserver#62 has been merged so you can write something like:
```python
oaiset = OAISet.query.get(get_oaiset_spec(self.id))
if oaiset is None:
    # create one or log error
else:
    oaiset.add_record(record)
```

krzysztof:
What is missing in this PR is the ckeditor...

PR #64, opened 2016-08-16
models: rename foreign keys

### LINKED ISSUE #63: models: rename foreign keys from `id_*` to `*_id`
- following discussion in https://github.com/inveniosoftware/invenio-groups/issues/39#issuecomment-238219783
- addresses #61 
## 

cc @lnielsen

alizeepace:
closes #63

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -39,3 +39,4 @@ Invenio module that adds support for communities.
 - Sami Hiltunen <sami.mikael.hiltunen@cern.ch>
 - Tibor Simko <tibor.simko@cern.ch>
 - Yoan Blanc <yoan.blanc@cern.ch>
+- Alizee Pace <alizee.pace@gmail.com>
--- a/invenio_communities/admin.py
+++ b/invenio_communities/admin.py
@@ -79,10 +79,10 @@ class InclusionRequestModelView(ModelView):
     can_delete = True
     can_view_details = True
     column_list = (
-        'id_community',
-        'id_record',
+        'community_id',
+        'record_id',
         'expires_at',
-        'id_user'
+        'user_id'
     )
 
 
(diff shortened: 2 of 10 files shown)


lnielsen:
I will merge this PR, as if merged now it will break Zenodo, so it has to wait until after Zenodo launch.

jirikuncar:
@lnielsen can you pin the Invenio-Community version on Zenodo?

lnielsen:
@jirikuncar No, since if we need to fix a problem, then we're forced to fix all the foreign keys as well. Zenodo is the only service using Communities module now, so the PR can stick around until we're ready.

Only alternative is to create a ``next`` branch and merge it there.

@alizeepace Just to be clear, there is no problem with the changes. My comments are just about timing when to apply them.

jirikuncar:
@lnielsen since Zenodo is out, can we merge this PR?

lnielsen:
@jirikuncar We're still fixing lots of bugs, and I prefer if we can wait with the update until we're out of bug fixing mode. Alternatively we can create a 1.0.0 final release with the current state, so that there's a maintenance branch that can receive fixes and that we can make releases from.

tiborsimko:
I wouldn't create any `maint-1.0` for an alpha package. It would be akin to sending mixed signals. The usual meaning we are trying to advertise is:

- alpha status: fast pace, don't look, no promises, API may change frequently, etc
- beta status: documentation exists, you can start looking at the package, but no data migration is provided 
- production status: data migration is fully supported, M.N.0 is released , `maint-M.N` is created

We could take this example to start musing about data migration related to table changes, but I'd say it's probably not worth it yet.  So I'd vote for option 2, merging whenever Zenodo is ready.

tiborsimko:
(That said, we'll face this situation sooner or later with other packages, so we should start thinking of taking *something* for an easy first data migration use case... We won't be able to postpone for much longer. But I'd say let's targe another, more stable package example, which may be perhaps more appropriate.  The communities is still young. See also related musings in https://github.com/inveniosoftware/invenio-github/pull/30#issuecomment-248859188)

jirikuncar:
@lnielsen it would be nice to do a release with the initial alembic recipes before integrating this PR. I have added a test case especially for the latest upgrade script so it should be good for a final review.

lnielsen:
Closing as Invenio-Communities is being completely revamped.

PR #69, opened 2016-09-02
global: addition of upload URL

slint:
Signed-off-by: Alexander Ioannidis <a.ioannidis@cern.ch>

### DIFF
--- a/invenio_communities/models.py
+++ b/invenio_communities/models.py
@@ -398,7 +398,7 @@ class Community(db.Model, Timestamp):
     @property
     def upload_url(self):
         """Get provisional URL."""
-        return 'TODO'
+        return url_for('invenio_deposit_ui.new', c=self.id, _external=True)
 
     @property
     def oaiset_spec(self):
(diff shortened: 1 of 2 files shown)


invenio_communities/models.py
```diff
@@ -398,7 +398,7 @@ def community_provisional_url(self):
     @property
     def upload_url(self):
         """Get provisional URL."""
-        return 'TODO'
+        return url_for('invenio_deposit_ui.new', c=self.id, _external=True)
```
jirikuncar: We should explicitly list `invenio-deposit` in `setup.py`.

invenio_communities/templates/invenio_communities/portalbox_upload.html
```diff
@@ -20,7 +20,7 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 #}
-<a href="TODO url for upload with community id" class="btn btn-primary btn-lg signup pull-right">Upload</a>
+<a href="{{ community.upload_url }}" class="btn btn-primary btn-lg signup pull-right">Upload</a>
```
jirikuncar: If we don't list Invenio-Deposit in setup.py then we should at least try/catch the build url error and return None and guard this section with `{% if community.upload_url %}`.

PR #72, opened 2016-09-06
global: logo file size fix

krzysztof:
* Adds logo size passing to ObjectVersion.

* Enables SVG support.

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -35,8 +35,11 @@ COMMUNITIES_REQUEST_EXPIRY_TIME = timedelta(days=365)
 COMMUNITIES_DELETE_HOLDOUT_TIME = timedelta(days=365)
 """Time after which the communities marked for deletion are hard-deleted."""
 
-COMMUNITIES_LOGO_EXTENSIONS = ['png', 'jpg', 'jpeg']
+COMMUNITIES_LOGO_EXTENSIONS = ['png', 'jpg', 'jpeg', 'svg']
 """Allowed file extensions for the communities logo."""
+
+COMMUNITIES_LOGO_MAX_SIZE = 1000 * 1000 * 1.5  # 1.5 MB
+"""Allowed file size for the communities logo."""
 
 COMMUNITIES_RECORD_KEY = 'communities'
 """Key inside the JSON record for communities."""
--- a/invenio_communities/forms.py
+++ b/invenio_communities/forms.py
@@ -121,9 +121,9 @@ class CommunityForm(Form):
 
     page = TextAreaField(
         description=_(
-            'Optional. A long description of the community collection,'
-            ' which will be displayed on a separate page linked from'
-            ' the index page.'),
+            'Optional. A long description of the community collection, '
+            'which will be displayed on a separate page linked from '
+            'the index page.'),
     )
 
     field_icons = {
@@ -136,9 +136,9 @@ class CommunityForm(Form):
     logo = FileField(
         label=_('Logo'),
         description=_(
-            'Optional. Image file used to aid and promote instant public'
-            ' recognition. Supported formats: png and jpg.'
-            ' Max file size: 1.5MB')
+            'Optional. Image file used to aid and promote instant public '
+            'recognition. Supported formats: PNG, JPG and SVG. '
+            'Max file size: 1.5 MB')
     )
 
     #
--- a/invenio_communities/utils.py
+++ b/invenio_communities/utils.py
@@ -26,6 +26,7 @@
 from __future__ import absolute_import, print_function
 
 import os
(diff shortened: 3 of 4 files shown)


krzysztof:
@lnielsen ping

invenio_communities/utils.py
```diff
@@ -99,13 +100,21 @@ def save_and_validate_logo(logo_stream, logo_filename, community_id):
     cfg = current_app.config
 
     logos_bucket_id = cfg['COMMUNITIES_BUCKET_UUID']
+    logo_max_size = cfg['COMMUNITIES_LOGO_MAX_SIZE']
     logos_bucket = Bucket.query.get(logos_bucket_id)
     ext = os.path.splitext(logo_filename)[1]
     ext = ext[1:] if ext.startswith('.') else ext
 
+    logo_stream.seek(SEEK_SET, SEEK_END)  # Seek from beginning to end
```
krzysztof: This is safe, since MAX_CONTENT_LENGTH will not allow anything more than 16MiB by this point anyway.

invenio_communities/config.py
```diff
@@ -35,9 +35,12 @@
 COMMUNITIES_DELETE_HOLDOUT_TIME = timedelta(days=365)
 """Time after which the communities marked for deletion are hard-deleted."""
 
-COMMUNITIES_LOGO_EXTENSIONS = ['png', 'jpg', 'jpeg']
+COMMUNITIES_LOGO_EXTENSIONS = ['png', 'jpg', 'jpeg', 'svg']
 """Allowed file extensions for the communities logo."""
 
+COMMUNITIES_LOGO_MAX_SIZE = 1000 * 1000 * 1.5  # 1.5 MB
```
jirikuncar: 1000 or 1024?

invenio_communities/forms.py
```diff
@@ -136,9 +136,9 @@ def has_autocomplete(self, field):
     logo = FileField(
         label=_('Logo'),
         description=_(
-            'Optional. Image file used to aid and promote instant public'
-            ' recognition. Supported formats: png and jpg.'
-            ' Max file size: 1.5MB')
+            "Optional. Image file used to aid and promote instant public"
```
jirikuncar: Why are you changing the quotes?

invenio_communities/config.py
jirikuncar: IMHO it should be in bytes: `1024 * 1024 * 1.5  # 1.5 Mb`

invenio_communities/views/ui.py
```diff
@@ -193,7 +193,7 @@ def new():
             if not community.save_logo(file.stream, file.filename):
                 form.logo.errors.append(_(
                     'Cannot add this file as a logo. Supported formats: '
-                    '  png and jpg. Max file size: 1.5MB.'))
+                    '  PNG, JPG and SVG. Max file size: 1.5MB.'))
```
jirikuncar: There are too many spaces before PNG.

invenio_communities/views/ui.py
```diff
@@ -234,7 +234,7 @@ def edit(community):
...
```
jirikuncar: There are too many spaces before PNG.

invenio_communities/config.py
krzysztof: ``1000 * 1000 * 1.5`` is still valid number number wrt to octets.

If we change it to ``1024^2 * 1.5`` then we should specify (in UI and errors) that the max file size is ``1.5 MiB`` which is less "common" unit than ``1.5 MB``.

invenio_communities/forms.py
krzysztof: Sorry, mistake.

invenio_communities/config.py
jirikuncar: > which is less "common" unit ...

That might be true, but it would be nice if defaults were in the units used in code (https://github.com/inveniosoftware/invenio-communities/pull/72#discussion_r77604622).

It might be more confusing for a user to see 1.5 MB but not being able to upload it:

```console
$ du -h output.dat
1.5M    output.dat
```

krzysztof: Yes, but it would be just his wrong assumption about what is 1.5 MB ;)
```
$ du output.dat -h                                                                                                                                                                                                              
1.5M	output.dat
                                                                                                                                                                                                                                   
$ du output.dat -h --si                                                                                                                                                                                                         
1.6M	output.dat
```

I understand the benefits of having size definitions in base 2 on the technical side of the application, but this is a user-facing size definition, and almost all user-facing size definitions in services like Dropbox, Google Drive etc. use SI units. I would argue that using "MiB" on the front-end is more confusing than the corner case you have in mind.

TDB tomorrow.

PR #76, opened 2016-09-23
global: communities accept/remove bugfix

krzysztof:
* Fixes the issue with accepting and removing communities raising assertion
  errors on external modification of 'communities' key in record.

### DIFF
--- a/invenio_communities/models.py
+++ b/invenio_communities/models.py
@@ -57,7 +57,7 @@ class InclusionRequest(db.Model, Timestamp):
 
     id_community = db.Column(
         db.String(100),
-        db.ForeignKey("communities_community.id"),
+        db.ForeignKey('communities_community.id'),
         primary_key=True
     )
     """Id of the community to which the record is applying."""
@@ -213,7 +213,7 @@ class Community(db.Model, Timestamp):
 
     def __repr__(self):
         """String representation of the community object."""
-        return "<Community, ID: {}>".format(self.id)
+        return '<Community, ID: {}>'.format(self.id)
 
     @classmethod
     def create(cls, community_id, user_id, **data):
@@ -267,9 +267,9 @@ class Community(db.Model, Timestamp):
 
         if p:
             query = query.filter(db.or_(
-                cls.id.like("%" + p + "%"),
-                cls.title.like("%" + p + "%"),
-                cls.description.like("%" + p + "%"),
+                cls.id.like('%' + p + '%'),
+                cls.title.like('%' + p + '%'),
+                cls.description.like('%' + p + '%'),
             ))
 
         if so in current_app.config['COMMUNITIES_SORTING_OPTIONS']:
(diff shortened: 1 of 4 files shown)


krzysztof:
Requires https://github.com/inveniosoftware/invenio-oaiserver/pull/96 and travis restart

invenio_communities/models.py
```diff
@@ -288,31 +288,35 @@ def add_record(self, record):
         key = current_app.config['COMMUNITIES_RECORD_KEY']
         record.setdefault(key, [])
 
-        assert self.id not in record[key]
-        record[key].append(self.id)
-        record[key] = sorted(record[key])
-
+        if self.has_record(record):
+            current_app.logger.warning(
```
lnielsen: ``warning -> info``?

invenio_communities/models.py
```diff
@@ -411,10 +415,23 @@ def oaiset_spec(self):
...
+        """Return the corresponding OAISet for given community.
+
+        :returns: returns OAISet object corresponding to this community.
+        :rtype: `invenio_oaiserver.models.OAISet`
+        """
+        if current_app.config["COMMUNITIES_OAI_ENABLED"]:
+            from invenio_oaiserver.models import OAISet
+            return OAISet.query.filter_by(spec=self.oaiset_spec).one()
+        else:
+            raise Exception("OAI is disabled for InvenioCommunities.")
```
lnielsen: ``return None`` instead?

invenio_communities/models.py
krzysztof: I'd keep it as warning, although this fallback is "safe", it should still indicate that something elsewhere is modifying record outside the API, which is most likely a bug or misuse.

invenio_communities/models.py
krzysztof: perhaps None is better, should at least log a warning?

jirikuncar: What about using SQLAlchemy `relationship`?
```python
# check if invenio_oaiserver is installed
Community.oaiset = relationship(OAISet, primary=Community.oaiset_spec==OAISet.spec, view_only=True)
```

invenio_communities/models.py
```diff
@@ -411,10 +415,23 @@ def oaiset_spec(self):
             community_id=self.id)
 
     @property
+    def oaiset(self):
+        """Return the corresponding OAISet for given community.
+
+        :returns: returns OAISet object corresponding to this community.
+        :rtype: `invenio_oaiserver.models.OAISet`
+        """
+        if current_app.config["COMMUNITIES_OAI_ENABLED"]:
```
jirikuncar: Can you try to use consistently single and double quotes?

invenio_communities/models.py
krzysztof: Where would you check if the oaiserver is installed?

jirikuncar: `pkg_resources.get_distribution('invenio-oaiserver')` and check `DistributionNotFound`

krzysztof: Hmm, I like the idea but since it's not clear where to inject this (outside model.py would be confusing) I'll push with ``None`` as return value.

krzysztof: @jirikuncar But that's why this flag came about as to not use this all over the place I suppose: https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/config.py#L67-L71

jirikuncar: IMHO it's just as a default value. One can still override it.

krzysztof: Almost there...
There's some issues with oaiset_spec being a property:
```
    Community.oaiset = db.relationship(
        OAISet, primaryjoin=OAISet.spec==Community.oaiset_spec, viewonly=True)
```
gives
```
sqlalchemy.exc.InvalidRequestError: One or more mappers failed to initialize - can't proceed with initialization of other mappers.  Original exception was: Could not locate any relevant foreign key columns for primary join condition 'oaiserver_set.spec = :spec_1' on relationship Community.oaiset.  Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or are annotated in the join condition with the foreign() annotation.
```
any ideas?

PR #85, opened 2016-10-18
models: UUID as community identifier

### LINKED ISSUE #84: models: use UUID for community ID and move the current ID to a separate field
As discussed IRL with @lnielsen and during DR meeting.
## Why:

**Reason 1:**
Current design require that every reference needs to be updated when the community is renamed.

**Reason 2**
B2Share would like to use this module. It is quite probable that it will have to support the migration of a community from one server to another. In order to avoid conflicts it would be needed to have unique identifiers for Communities.

**Reason 3**
B2Share builds endpoints on top of the communities REST API endpoint. One of those endpoints is the communities' schemas /api/communities/<ID>/schemas/<version>.
If the community is renamed every record cannot validate anymore. We could update the current records but all the previous versions become hard to work with.

**Reason to keep a user friendly id in a separate field:**
Zenodo has an endpoint for each community. Thus it needs a unique human friendly name for each community.
## Who needs it:
- Zenodo for reason 1
- B2Share for reasons 1 and 2

nharraud:
- NEW Renames previous community ID field as "name" and creates a new "id"
  field which is an UUID. UI endpoints still use the name, REST API
  endpoints use the new ID. (closes #84)

Signed-off-by: Nicolas Harraudeau nicolas.harraudeau@cern.ch

### DIFF
--- a/invenio_communities/models.py
+++ b/invenio_communities/models.py
@@ -27,6 +27,8 @@
 from __future__ import absolute_import, print_function
 
 import hashlib
+import uuid
+from copy import deepcopy
 from datetime import datetime
 
 from flask import current_app, url_for
@@ -56,7 +58,7 @@ class InclusionRequest(db.Model, Timestamp):
     __tablename__ = 'communities_community_record'
 
     id_community = db.Column(
-        db.String(100),
+        UUIDType,
         db.ForeignKey('communities_community.id'),
         primary_key=True
     )
@@ -166,8 +168,18 @@ class Community(db.Model, Timestamp):
 
     __tablename__ = 'communities_community'
 
-    id = db.Column(db.String(100), primary_key=True)
+    id = db.Column(
+        UUIDType,
+        primary_key=True,
+        default=uuid.uuid4,
+    )
     """Id of the community."""
+
+    name = db.Column(
+        db.String(100),
+        unique=True
+    )
+    """Human readable identifier of the community."""
 
     id_user = db.Column(
         db.Integer,
@@ -216,10 +228,14 @@ class Community(db.Model, Timestamp):
         return '<Community, ID: {}>'.format(self.id)
 
     @classmethod
-    def create(cls, community_id, user_id, **data):
+    def create(cls, community_name, user_id, community_id=None, **data):
         """Get a community."""
+        if community_id:
+            data = deepcopy(data)
+            data['id'] = community_id
         with db.session.begin_nested():
-            obj = cls(id=community_id, id_user=user_id, **data)
+            obj = cls(name=community_name,
+                      id_user=user_id, **data)
             db.session.add(obj)
         return obj
 
@@ -232,9 +248,15 @@ class Community(db.Model, Timestamp):
         return False
 
     @classmethod
-    def get(cls, community_id, with_deleted=False):
+    def get(cls, community_id=None, community_name=None, with_deleted=False):
         """Get a community."""
-        q = cls.query.filter_by(id=community_id)
+        assert((community_id is not None) ^ (community_name is not None),
+               'One and only one of community_id or community_name '
+               'should be != None.')
+        if community_id:
+            q = cls.query.filter_by(id=community_id)
+        else:
+            q = cls.query.filter_by(name=community_name)
         if not with_deleted:
             q = q.filter(cls.deleted_at.is_(None))
         return q.one_or_none()
@@ -267,7 +289,7 @@ class Community(db.Model, Timestamp):
 
         if p:
             query = query.filter(db.or_(
-                cls.id.like('%' + p + '%'),
+                cls.name.like('%' + p + '%'),
                 cls.title.like('%' + p + '%'),
                 cls.description.like('%' + p + '%'),
             ))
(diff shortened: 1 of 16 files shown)


nharraud:
@lnielsen The main change for the users is in the API as it now uses the UUID both in the endpoint and the records' metadata. Would this be ok for Zenodo?

jirikuncar:
@nharraud how does it play together with #64?

invenio_communities/models.py
```diff
@@ -240,6 +256,20 @@ def get(cls, community_id, with_deleted=False):
         return q.one_or_none()
 
     @classmethod
+    def get(cls, community_id=None, community_name=None, with_deleted=False):
+        """Get a community."""
+        assert((community_id is not None) ^ (community_name is not None),
```
jirikuncar: Why not both?

```
q = cls.query
if community_id is not None:
    q = q.filter_by(id=community_id)
if community_ name is not None:
    q = q.filter_by(name =community_ name)
```

nharraud:
@jirikuncar I didn't test it with #64 yet. I first wanted to have your and @lnielsen feedback on the API and model changes.

invenio_communities/models.py
jirikuncar: IMHO there is no need for the `get` wrapper at this point.

jirikuncar: Moreover there is one defined on line 235:`     def get(cls, community_id, with_deleted=False):`

nharraud: Sorry, I thought I had removed it, this one is just extending the previous implementation.

jirikuncar: If they are both unique then you can do `db.or_(Community.id == community_id, Comunity.name == community_id))` and keep the API simple.

invenio_communities/models.py
```diff
@@ -232,9 +248,15 @@ def save_logo(self, stream, filename):
         return False
 
     @classmethod
-    def get(cls, community_id, with_deleted=False):
+    def get(cls, community_id=None, community_name=None, with_deleted=False):
         """Get a community."""
-        q = cls.query.filter_by(id=community_id)
+        assert((community_id is not None) ^ (community_name is not None),
+               'One and only one of community_id or community_name '
```
krzysztof: I usually use something like "Provide eitehr X or Y but not both." as error.
It's irrelevant to user of this function that ``community_name`` or ``community_id`` is ``None`` by default and we do some bitwise tricks to get logical XOR on parameters.

krzysztof:
@nharraud Sorry, I might have missed some RFC at some point - why is this change required?

invenio_communities/models.py
nharraud: Good point.

invenio_communities/models.py
nharraud: Wouldn't this make the API even more strange as this gets multiple communities when the community_name and community_id don't match the same Communities?
If I am not mistaken every other `get` method in Invenio APIs returns one object max.

nharraud:
@krzysztof You can see the requirement in the issue this PR closes.

tests/test_invenio_communities.py
```diff
@@ -284,34 +284,39 @@ def assert_header_links(response, ref, page, size):
 
 def test_communities_rest_get_details(app, db, communities):
     """Test the OAI-PMH Sets creation."""
+    (comm1, comm2, comm3) = communities
     with app.test_client() as client:
-        response = client.get('/api/communities/comm1')
+        response = client.get('/api/communities/{}'.format(str(comm1.id)))
```
krzysztof: So the endpoint and serialization would have UUID instead of name from now on?

nharraud: only the API yes. As the UUIDs are written inside the records, one needs to retrieve the community info using the UUID.
It's also possible to allow retrieving by either name or UUID as both are unique. We would just need to make sure that names which look like UUIDs are not allowed. Is the communities API part of the API currently documented on Zenodo? If yes then we will need to keep it backward compatible.

krzysztof:
If we keep list of UUIDs in record["communities"] instead of community names it kind of defeats the purpose of having this key in the record's JSON in the first place. AFAIR, the argument for "communities" being a key in the record instead of ORM M2M table was that it is a valid record metadata, relevant outside of the invenio instance. If this was to become a UUID, I don't see a reason for keeping this key in the record anymore. For REST responses we would have to serialize it back to names, similarly for elasticseach.

At this point this stops being any useful metadata and should be modeled in ORM, and the serialized accordingly for mapping and export formats.

tests/test_invenio_communities.py
krzysztof: Communities API is not part of Zenodo documentation at the moment, however we might want to add it at some point. However, Deposit API is already using community names (I don't think we would want to change that), hence we might want to stick with names instead of UUIDs across all of our APIs relating to communities.

jirikuncar: What about moving to a solution using PIDStore? We can take advantage of redirects, stub pages, etc. We could still keep the API using community name, but it would be a property with registered PID value.

jirikuncar: IMHO we should keep the registered PID (community name) inside the record so for Zenodo nothing should change and thanks to the redirects it should be fine also for B2Share use-case.

nharraud: This would still keep the the issue @lnielsen raised, i.e. having a celery task renaming every record when a community name changes. This wouldn't be an issue if this wasn't creating potentially thousands of row in the database as they are versioned.

jirikuncar: > This wouldn't be an issue if this wasn't creating potentially thousands of row in the database as they are versioned.

This should not influence our architectural decision here. Space is cheap and community should not be renamed everyday.

slint:
Closed since the module is currently being refactored

PR #96, opened 2017-01-17
models: add cascade to all foreign keys

david-caro:
Signed-off-by: David Caro <david@dcaro.es>


jirikuncar:
We should include a upgrade recipe and see with #64.

david-caro:
I don't have much time right now, so it might take me some time to get the upgrade recipe, I can use that pr as base for this one (as probably will get merged before too). Thanks for the pointer ;) (looks also as a good example of upgrade recipe).

slint:
Closed since the module is being refactores

PR #98, opened 2017-02-18
bundles: addition of `cleancssurl` filter

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>

### DIFF
--- a/invenio_communities/bundles.py
+++ b/invenio_communities/bundles.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2017 CERN.
 #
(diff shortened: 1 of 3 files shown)


invenio_communities/bundles.py
```diff
@@ -42,7 +42,7 @@
 
 css = NpmBundle(
     'scss/invenio_communities/communities.scss',
-    filters='scss, cleancss',
+    filters='scss,cleancssurl',
```
jmartinm: Is this needed after https://github.com/miracle2k/webassets/pull/486 was merged? Or just using updated `webassets` is enough?

drjova: Using updated ``webassets`` should be enough :D

lnielsen: yep webassets is enough **once it is released**

slint:
Closed since we're moving away from webassets

Issue #99, opened 2017-02-20
api: case insensitive search


slint:
Communities filtering uses the regular case-sensitive `like` clause in the [search query](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/models.py#L273-L275). Using the case-insensitive `ilike` is probably more appropriate in order to emulate Elasticsearch behavior.

krzysztof:
Closed by #100

PR #100, opened 2017-03-02
Text search: case insensitive search

jainaman224:
Made text search case insensitive and matches word instead of substring.

Added test cases to test filter_community function in Community class.

#99

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Model functions tests."""
+
+from __future__ import absolute_import, print_function
+
+import pytest
+
+from invenio_communities.models import Community
+
+
+@pytest.mark.parametrize('case_modifier', [
+    (lambda x: x),
+    (lambda x: x.upper()),
+    (lambda x: x[0].upper() + x[1:]),
+])
+def test_filter_community(app, db, communities_for_filtering, user,
+                          case_modifier):
+    """Test the community filter task."""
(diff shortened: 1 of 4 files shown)


tests/test_models.py
```diff
@@ -0,0 +1,73 @@
...
+    with app.app_context():
+        (comm0, comm1, comm2) = communities_for_filtering
+        rec1 = Record.create({
+            'title': 'Foobar and Bazbar',
+            'description': 'On Foobar, Bazbar and <b>more</b>.'
+        })
+        InclusionRequest.create(community=comm0, record=rec1, user=user)
+
+        # Case insensitive
+        results = Community.filter_communities(p='wide learning',
```
krzysztof: This test looks really good already, but since the original issue was about case insensitivity, add some tests with Upcase and UPPERCASE query strings and show that they work the same. You can do that nicely with parametrized tests (http://doc.pytest.org/en/latest/parametrize.html)
Something like:

```python
@pytest.mark.parametrize("casemodifier", [
    (lambda x: x),
    (lambda x: x.upper()),
    (lambda x: x[0].upper() + x[1:]),
])
```

tests/test_models.py
```diff
@@ -0,0 +1,83 @@
...
+        (comm0, comm1, comm2) = communities_for_filtering
+        rec1 = Record.create({
+            'title': 'Foobar and Bazbar',
+            'description': 'On Foobar, Bazbar and <b>more</b>.'
+        })
+        InclusionRequest.create(community=comm0, record=rec1, user=user)
+
+        # Case insensitive
+        def test_case_insensitivity(search_input):
+            results = Community.filter_communities(p=case_insensitive,
```
krzysztof: This code never runs since you don't use this function anywhere.

krzysztof: Also since you wrap this test in a ``parametrize`` decorator, the whole test is repeated four times, hence all assertions below line 60 are also repeated four times (even though they don't depend on the ``case_insensitive`` parameter at all).

tests/test_models.py
```diff
@@ -0,0 +1,83 @@
...
+                          case_insensitive):
+    """Test the community filter task."""
+
+    with app.app_context():
+        (comm0, comm1, comm2) = communities_for_filtering
+        rec1 = Record.create({
+            'title': 'Foobar and Bazbar',
+            'description': 'On Foobar, Bazbar and <b>more</b>.'
+        })
+        InclusionRequest.create(community=comm0, record=rec1, user=user)
```
krzysztof: You don't need to create a record nor the inclusion request. You are only creating communities, and only searching for Community objects (not their records).

tests/test_models.py
```diff
@@ -0,0 +1,83 @@
...
+    ('Beautiful'),
+    ('beautiful'),
+    ('bEAutiful'),
+    ('BEAUTIFUL'),
+])
+def test_filter_community(app, db, communities_for_filtering, user,
+                          case_insensitive):
+    """Test the community filter task."""
+
+    with app.app_context():
```
krzysztof: if you use the ``app`` fixture in as parameter, you don't have to put this code inside ``app.app_context()``, since it's already provided: https://github.com/inveniosoftware/invenio-communities/blob/master/tests/conftest.py#L100-L101

slint:
For tests to stop failing, updating Sphinx to `v1.5.1` in [`setup.py`](https://github.com/inveniosoftware/invenio-communities/blob/master/setup.py#L58) will do it.

(Fix based on https://github.com/inveniosoftware/cookiecutter-invenio-module/commit/a561e351ca2924dc657bf6a166f6780fd4fd335e)

PR #106, opened 2017-06-15
api: addition of community created updated timestamps

krzysztof:


invenio_communities/serializers/schemas/community.py
```diff
@@ -41,6 +41,8 @@ class CommunitySchemaV1(Schema):
     page = fields.String()
     curation_policy = fields.String()
     last_record_accepted = fields.DateTime()
+    created = fields.DateTime()
```
lnielsen: See https://sqlalchemy-utils.readthedocs.io/en/latest/models.html#module-sqlalchemy_utils.models

PR #107, opened 2017-06-17
addition of community id_user to REST API serialization

krzysztof:


krzysztof:
@lnielsen is it OK to add this?

PR #109, opened 2017-10-23
views: new community form action fix

slint:
Signed-off-by: Alexander Ioannidis <a.ioannidis@cern.ch>


lnielsen:
Do you need a release? if so just submit a PR

PR #121, opened 2019-05-10
setup: include newer versions of ES

Glignos:


slint:
Closed in favor of #123

PR #127, opened 2020-01-13
demo: add communities

frankois:


slint:
This will eventually be added in InvenioRDM to be able to run through invenio-cli demo .... The code has been extracted into a GitHub gist: https://gist.github.com/slint/dc7edc04bae84213e4b1588e95486e5a

PR #134, opened 2020-01-14
WIP: initial implementation of members

Glignos:


invenio_communities/views.py
```diff
@@ -0,0 +1,117 @@
...
+            'pid(comid,record_class="invenio_communities.api:Community",'
+            'object_type="com")'),
+        view_func=comm_view,
+    )
+
+    request_view = MembershipRequest.as_view(
+        'community_requests_api'
+    )
+    community_members_rest_blueprint.add_url_rule(
+        '/communities/requests/<token>',
```
slint: Since this is the REST API, it should follow the resources model, i.e. expose:

-  `(GET|POST) /api/communities/requests/`
- `(GET|PUT|DELETE) /api/communities/requests/<request-id>`

The UI view `/communities/requests?token=<...>` is responsible for parsing/validating the token, and rendering the appropriate accept/reject box with the resolved `request-id` injected in the Jinja (to be used by the Javascript app to make a `POST /api/communities/requests/<request-id>` request). Let's discuss IRL though, because there are some tricky details on how auth would work

invenio_communities/views.py
```diff
@@ -0,0 +1,117 @@
...
+from flask import Blueprint
+from invenio_communities.api import CommunityMembersCls, MembershipRequestCls
+from flask_security import current_user
+
+def create_blueprint_from_app(app):
+    community_members_rest_blueprint = Blueprint(
+        'invenio_communities',
+        __name__,
+        url_prefix="",
+        template_folder="./templates"
```
slint:
```suggestion
        template_folder="templates"
```

invenio_communities/views.py
```diff
@@ -0,0 +1,117 @@
...
+
+from flask import Blueprint
+from invenio_communities.api import CommunityMembersCls, MembershipRequestCls
+from flask_security import current_user
+
+def create_blueprint_from_app(app):
+    community_members_rest_blueprint = Blueprint(
+        'invenio_communities',
+        __name__,
+        url_prefix="",
```
slint:
```suggestion
```

invenio_communities/views.py
```diff
@@ -0,0 +1,117 @@
...
+            location='json',
+            required=True,
+            validate=[validate.OneOf(['invitation', 'request'])]
+        )
+
+    }
+    # TODO: change invite_type name
+    @use_kwargs(post_args)
+    def post(self, email=None, pid_value=None, role=None, invite_type=None):
+        pid, community = pid_value.data
```
slint: You can also use a `pass_community` decorator (similar to [`pass_record` from `invenio-records-rest`](https://github.com/inveniosoftware/invenio-records-rest/blob/4c98abfc99132e02b315855da018499694cc7227/invenio_records_rest/views.py#L373-L388)), to pass these as `comid=None` and `community=None`.

invenio_communities/views.py
```diff
@@ -0,0 +1,117 @@
...
+        if invite_type == 'invitation':
```
slint: This can be inferred from if the `current_user` is already part of the community or not:

- `current_user.id in community.members -> "invitation"`
- `current_user.id not in community.members -> "request"`

invenio_communities/api.py
```diff
@@ -23,12 +30,76 @@ class Community(Record):
...
+            else:
+                self.model.delete()
+
+        return self
+
+
+class CommunityMembersCls(object):
+
+    @classmethod
+    def invite_member(cls, community, email, role):
```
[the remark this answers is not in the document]
Glignos: The reasoning is that since there is app state logic e.g. fetching logged in user, I preferred to keep the model functions clean of this and have them interact only with the DB.

invenio_communities/views.py
Glignos: This is not very trustworthy. It's a stretch but imagine the following scenario.

1. I send an email to one of my email addresses.
2. Then I am removed from the community.
3. I then click on the link and accept the invitation to rejoin it.

If I'm not mistaken something like this could then be possible.
I understand that it is a bit complicated currently, so we can also talk IRL for a better solution.

invenio_communities/models.py
```diff
@@ -10,12 +10,203 @@
...
+    role = db.Column(
+        ChoiceType(CommunityRoles, impl=db.CHAR(1)), nullable=False)
+
+    @property
+    def email(self):
+        """Get user email."""
+        return User.query.get(self.user_id).email
+
+    @classmethod
+    def create_or_modify(cls, membership_request):
```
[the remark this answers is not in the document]
Glignos: I think instead I will use a `PUT` method only for the modification and have it more explicit.

PR #159, opened 2020-05-27
assets: use invenio-search-ui for search

Glignos:


invenio_communities/assets/semantic-ui/js/invenio_communities/search.js
```diff
@@ -0,0 +1,94 @@
...
+        </Item.Description>
+      </Item.Content>
+    </Item>
+  );
+}
+
+overrideStore.add("ResultsList.item", ResultsItemTemplate);
+overrideStore.add("ResultsGrid.item", ResultsGridItemTemplate);
+
+const searchConfig = {
```
lnielsen, line 41: should this be injected from Flask config similar to done in Invenio-Search-UI?

slint: I agree, and we discussed with @Glignos that the general mechanism provided by default has to be changed in invenio-search-ui, to allow for extra/hidden parameters, etc. 

The plan, for now, is to get the current PR integrated, so that we have something working with the new `invenio-search-ui` features, and then incrementally improve the "unstable" helpers. We have a list of improvements that we'll convert into issues in `invenio-search-ui`.

PR #169, opened 2020-06-02
fix various styling issues

### LINKED ISSUE #167: theme: fix header search
The header search bar is not consistent

image

Should be like this:

image

zzacharo:
* Fixes header search bar in search and records search
* Fixes community padding from header

closes #167 

Screenshot 2020-06-02 at 11 01 39

Screenshot 2020-06-02 at 11 40 49

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/records_search.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/records_search.js
@@ -1,11 +1,11 @@
 import ReactDOM from "react-dom";
 import React from "react";
-import { Card, Item, Button } from "semantic-ui-react";
+import { Card, Input, Item } from "semantic-ui-react";
 import _truncate from "lodash/truncate";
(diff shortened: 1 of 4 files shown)


zzacharo:
@slint the community should not define a search bar as it uses the invenio-theme default template...The problem now is, that community is declaring new search pages and therefore we cannot override the searchbar from RDM. I think this problem needs to be addressed in https://github.com/inveniosoftware/react-searchkit/issues/117. Then, we will be able to override components per search application.

slint:
In the end, it should be using its own search bar (probably define using overridable), and the navbar shouldn't be part of the React application at all, like so:

image

zzacharo:
I think the problem is that invenio-theme defines [here]() the `header-search-bar` id and then if the invenio-search-ui is included then by default it renders the searchbar there. We need to think a way to make that optional in a easier way than overriding the `header` template :) If it's fine with you can I merge this one now and address that issue later on?

PR #171, opened 2020-06-04
Migrate form fields [WIP]

Glignos:


invenio_communities/assets/semantic-ui/js/invenio_communities/new.js
```diff
@@ -181,8 +166,10 @@ const CommunityCreateForm = () => {
             website: Yup.string().url("Must be a valid URL"),
             funding: Yup.array().of(Yup.string()
             .max(20, "Must be 20 characters or less")),
-            domain: Yup.array().of(Yup.string()
-            .max(20, "Must be 20 characters or less")),
+            domain: Yup.array().of(Yup.string().oneOf(
+              config.domains.map( c => {
+                return c.value;
+              }))),
```
slint, line 172: If possible add validation to the backend as well in the marshmallow schema

invenio_communities/assets/semantic-ui/js/invenio_communities/new.js
```diff
@@ -12,29 +12,33 @@ import * as Yup from "yup";
...
-import { DeleteActionButton, ArrayField, SelectField, StringField } from './fields'
-import { Icon, Grid, Container } from "semantic-ui-react";
+import { DeleteActionButton } from './fields'
+import { Icon, Grid } from "semantic-ui-react";
+import { SelectField, TextField } from 'react-invenio-forms';
 
 const domContainer = document.querySelector("#is_new_community");
 const __IS_NEW = JSON.parse(domContainer.dataset.config);
 
+const domContainer2 = document.querySelector("#form_config");
```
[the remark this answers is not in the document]
Glignos: If we move this way we should also change the id of the div to a more generic one. But I agree that this should make it more elegant.

invenio_communities/marshmallow/json.py
```diff
@@ -50,6 +54,32 @@ def serialize_creator(record, context):
...
+def validate_domain(domain, context):
+    """Load the record creator."""
+    if domain is missing:
+        return missing
+    else:
+        valid_domains = [
+            dom['value'] for dom
+            in LocalProxy(
+                lambda: current_app.config[
+                    'COMMUNITIES_FORM_CONFIG']['domains'])]
```
slint, line 66: `LocalProxy` shouldn't be necessary here, since `validate_domain` and the marshmallow schema should probably always be called inside an application context.

slint: Maybe though, you want to use something like `marshmallow.validate.OneOf`, in which case you should use a `LocalProxy` like so:

```python
from marshmallow import validate, Schema, fields

def domain_choices():
    domains = current_app.config['COMMUNITIES_FORM_CONFIG']['domains']
    return [d['value'] for d in domains]


class Community(Schema):

    domains = SanitizedUnicode(
        validate=validate.OneOf(LocalProxy(domain_choices)),
    )
```

invenio_communities/templates/semantic-ui/invenio_communities/settings.html
```diff
@@ -49,6 +49,9 @@
   <div class="thirteen wide column">
     <div id="app"></div>
   </div>
-  <div id="is_new_community" data-config='{{ False | tojson }}'></div>
+  <div
+    id="form_config"
+    data-is_new='{{ False | tojson }}'
```
slint, line 54:
```suggestion
    data-is-new='{{ False | tojson }}'
```

Which translates to `container.dataset.isNew`

PR #180, opened 2020-08-16
global: address TODOs

Glignos:
* Also changes Community Record URLs to be on par with the members
* Adds email response for community records requests


invenio_communities/records/api.py
```diff
@@ -245,13 +245,16 @@ def as_dict(self, include_requests=True):
         res = {
             **self,
             'status': str(self.status.title),
-            'record_pid': self.record.pid.pid_value,
-            'community_pid': self.community.pid.pid_value
+            'recid': self.record.pid.pid_value,
+            'comid': self.community.pid.pid_value,
+            'community_title': self.community['title'],
+            'record_title': self.record['titles'][0]['title'],
```
slint, line 251: This is specific to each record's metadata schema, which we cannot know in advance (i.e. we can put new things in the record, but cannot assume how other fields of the record look like). Also, this doesn't seem to be used anywhere.

tests/api/records/conftest.py
```diff
@@ -31,7 +31,7 @@ def record_owner(db):
 def record(db, es, record_owner):
     """Record fixture."""
     record = Record.create({
-        'title': 'Title',
+        'titles': [{'title': 'Test title', 'lang': 'eng'}],
```
slint, line 34: This is specific to Invenio-RDM-Records, so the simple title should be fine

invenio_communities/assets/semantic-ui/js/invenio_communities/records.js
```diff
@@ -282,7 +315,110 @@ const App = () => {
   );
 };
 
+const CommunitiesCollectionDropdown = ({ communityRecord }) => {
+  const [isOpen, setIsOpen] = useState(false);
```
slint, line 319: This one seems to be unused

invenio_communities/ext.py
```diff
@@ -65,11 +66,20 @@ def init_app(self, app):
...
+            def communities_ui_data(record):
                 from invenio_communities.records.api import Record, \
                     RecordCommunitiesCollection
-                return RecordCommunitiesCollection(Record(record, model=record.model))
-            return dict(record_communities=record_communities)
+                record_commmunities = RecordCommunitiesCollection(
+                    Record(record, model=record.model))
+                record_communities_dict = record_commmunities.as_dict()
+                communities_dict = {
+                    record_community.community['id']: record_community.community
```
slint, line 76: Probably already works since the `Community` is a dict subclass

```suggestion
                    record_community.community['id']: record_community.community.dumps()
```

Also not sure if `community['id']` is the comid or the UUID of the community (and it shouldn't be either actually, but `community.comid.pid_value` should be used probably)

invenio_communities/ext.py
```diff
@@ -65,11 +66,20 @@ def init_app(self, app):
         @app.context_processor
         def record_context_processors():
             """Jinja context processors for communities record integration."""
-            def record_communities(record):
+            def communities_ui_data(record):
```
slint, line 69: This name is pretty generic, even though the function ends up being used only in the record view's JS app.

slint (approved):
Some small naming comments and minor stuff, but otherwise LGTM

invenio_communities/token.py
```diff
@@ -16,84 +16,59 @@
 from flask import current_app
 from itsdangerous import BadData, JSONWebSignatureSerializer, SignatureExpired
 
-SUPPORTED_DIGEST_ALGORITHMS = ('HS256', 'HS512')
 
+def create_token(obj_id, extra_data={}):
```
slint, line 20:
```suggestion
def create_token(payload=None):
```

and `payload = payload or {}` in the body

(see comment below for why `obj_id` is removed)

invenio_communities/token.py
```diff
@@ -16,84 +16,59 @@
...
+    """Load data in a token.
+    :param token: Token to load.
+    :param force: Load token data even if signature expired.
+                    Default: False.
     """
+    jss = JSONWebSignatureSerializer(current_app.config['SECRET_KEY'])
+
+    try:
+        data = jss.loads(token)
+    except SignatureExpired as e:
```
slint, line 65: As discussed IRL, let's instead catch the superclass exception (BadSignature) and re-raise as our own custom HTTPException-based error (to have a proper 400 response returned)

invenio_communities/token.py
```diff
@@ -16,84 +16,59 @@
...
-                rnd=binascii.hexlify(os.urandom(4)).decode('utf-8')
-            )
+    Note random data is added to ensure that no two tokens are identical.
+    """
+    jss = JSONWebSignatureSerializer(current_app.config['SECRET_KEY'])
+    return jss.dumps(
+        dict(
+            id=obj_id,
+            data=extra_data,
+            rnd=binascii.hexlify(os.urandom(4)).decode('utf-8')
```
slint, line 30: Not sure if this is needed anymore.... Thus `expected_value` below can now check against the entire payload (instead of just the `data['id']`)

PR #187, opened 2020-09-23
Fix communities creation check

Glignos:


invenio_communities/templates/invenio_communities/mycommunities.html
```diff
@@ -10,10 +10,13 @@
...
 {% if current_user.is_authenticated %}
   {% if mycommunities %}
+  {% if not can_user_create_community(current_user) %}
+    <p class="alert alert-warning small">To create a community your account must be verified for at least one week. If you want to speed up the process, you can contact our support team describing your case.</p>
+  {% endif %}
   <h2>
     {{ _('My communities') }}
     <div class="pull-right">
-      &nbsp;<a href="{{url_for('.new')}}" class="btn btn-primary"><i class="icon-plus-sign"></i> {{ _('New') }}</a>
+      &nbsp;<a href="{{url_for('.new')}}" class="btn btn-primary" {{ "disabled" if not can_user_create_community(current_user) }}><i class="icon-plus-sign"></i> {{ _('New') }}</a>
```
slint, line 19:
```suggestion
      &nbsp;<a href="{{url_for('.new')}}" class="btn btn-primary {{ "disabled" if not can_user_create_community(current_user) }}" {{ "disabled" if not can_user_create_community(current_user) }}><i class="icon-plus-sign"></i> {{ _('New') }}</a>
```

This also makes the button unclickable

invenio_communities/templates/invenio_communities/mycommunities.html
```diff
@@ -10,10 +10,13 @@
 <div class="well">
 {% if current_user.is_authenticated %}
   {% if mycommunities %}
+  {% if not can_user_create_community(current_user) %}
```
slint, line 13: I would extract this as a variable at the beginning of the template:

```jinja
{% set can_create_community = can_user_create_community(current_user) %}
```

and then use appropriately below

invenio_communities/views/ui.py
```diff
@@ -177,6 +176,10 @@ def new():
             for at least one week. If you want to speed up the process, you \
```
slint, line 176: Instead of `\`, it's better to use string literal lines in parentheses, just to be sure about the whitespace

PR #218, opened 2021-04-15
communities: add skeleton for UI views

zzacharo:
Enabling the ui views for communities. More specifically the following pages exist (UI is not updated but just revived):
- [x] Communities index page
- [x] Communities search page
- [x] Communities details page
- [x] Communities creation page

# Screenshots

## Details page
Screenshot 2021-04-15 at 14 57 30

## Creation page
Screenshot 2021-04-15 at 14 57 44

## Index page
Screenshot 2021-04-15 at 14 57 04

## Search Page
Screenshot 2021-04-15 at 14 57 18


invenio_communities/views/communities.py
```diff
@@ -0,0 +1,47 @@
...
+from flask import render_template
+from flask_login import login_required
+
+from .decorators import pass_community
+
+#
+# Views
+#
+
+@login_required
```
lnielsen, line 20: Minor: I don't think we need to require login.

invenio_communities/views/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+from functools import wraps
+from flask import g
+from invenio_communities.proxies import current_communities
+
+
+def service():
+    """Get the record service."""
+    return current_communities.communities_service
```
lnielsen, line 15:
```suggestion
    return current_communities.service
```

invenio_communities/views/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def pass_community(f):
+    """Decorate to retrieve the community record using the community service.
+    """
+    @wraps(f)
+    def view(**kwargs):
+        pid_value = kwargs.get('pid_value')
+        record = service().read(
```
lnielsen, line 24: Comment: If we call everything a record I think it will get quite confusing in the end:
```suggestion
        community = service().read(
```

invenio_communities/views/decorators.py
zzacharo: @lnielsen I prefixed it because in the future we might have different services for members and requests, no? Or should we follow the `record_service.<subservice>` way?

Issue #235, opened 2021-04-19
global: community ID rename funcitonality


slint:
When changing the ID of the community, there might be several (expensive) operations that have to take place, especially after we integrate records.

- [x] Add service method `.rename()` to isolate the operation
- [x] `POST /api/communities/{:comid}/rename` endpoint
- [ ] Update the Profile settings form to call the above endpoint

PR #239, opened 2021-04-20
resources: adds renaming community ID endpoint

Glignos:
* Addresses https://github.com/inveniosoftware/invenio-communities/issues/235

### DIFF
--- a/invenio_communities/communities/permissions.py
+++ b/invenio_communities/communities/permissions.py
@@ -97,11 +97,6 @@ class CommunityPermissionPolicy(BasePermissionPolicy):
 
     can_delete = [CommunityOwners()]
 
-    can_search = [
-        IfRestricted(
-            'visibility',
-            then_=[CommunityOwners()],
-            else_=[AnyUser()]),
-        ]
+    can_search = [AnyUser()]
 
-    can_search_user_communities = [CommunityOwners()]
+    can_search_user_communities = [AuthenticatedUser()]
--- a/invenio_communities/communities/permissions.py
+++ b/invenio_communities/communities/permissions.py
@@ -1,106 +1,107 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020-2021 CERN.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Community permissions."""
 
+from elasticsearch_dsl.query import Q
 from flask_principal import UserNeed
-from flask_security import current_user
 from invenio_access.permissions import any_user
-from invenio_records_permissions.generators import AnyUser, Generator
+from invenio_rdm_records.services.generators import \
+    IfRestricted as IfRestrictedBase
+from invenio_records_permissions.generators import AnyUser, \
+    AuthenticatedUser, Generator
 from invenio_records_permissions.policies import BasePermissionPolicy
 
-# from invenio_communities.members.permissions import CommunityMember
 
+class IfRestricted(IfRestrictedBase):
+    """IfRestricted.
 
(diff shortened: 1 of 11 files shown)


invenio_communities/communities/service.py
```diff
@@ -10,10 +10,45 @@
...
+
+        # Permissions
+        self.require_permission(identity, "rename", record=record)
+
+        if 'id' not in data:
+            raise ValidationError('Missing data for required field.', 'id')
+
+        if record.pid.pid_value == data['id']:
+            raise ValidationError(
+                'A new ID value is required for the renaming', 'id')
```
slint, line 35: This can go in the component.

invenio_communities/communities/permissions.py
```diff
@@ -1,106 +1,107 @@
...
 
-class CommunityOwner(Generator):
-    """Returns community owner need."""
+    def needs(self, record=None, **kwargs):
+        """Enabling Needs."""
+        if record:
+            return [
+                UserNeed(owner.owner_id)
+                for owner in record.access.owned_by
+                ]
```
slint, line 71:
```suggestion
            ]
```

Shelved: I see some other places as well with weird indentation. I'm shelving this in a separate issue, so that we merge this guy though.

invenio_communities/communities/permissions.py
```diff
@@ -1,106 +1,107 @@
...
 
-
-class CommunityVisibilityPolicy(Generator):
-    """Allows only community members to read the community if its hidden."""
-
-    def needs(self, community=None, **kwargs):
-        """."""
-        if community['visibility'] == 'public':
-            return [any_user]
+class IfRestricted(IfRestrictedBase):
```
slint, line 21: Shelved: As discussed IRL, since we only need to override the `access.visibility.{private,public}` variable here, this could be extracted or made uniform with theoriginal class.

invenio_communities/views/api.py
```diff
@@ -9,7 +9,13 @@
...
+def create_communities_api_blueprint(app):
     """Create communities api blueprint."""
     ext = app.extensions["invenio-communities"]
     return ext.communities_resource.as_blueprint()
+
+
+def create_user_communities_api_blueprint(app):
+    """Create user communities api blueprint."""
+    ext = app.extensions["invenio-communities"]
+    return ext.user_communities_resource.as_blueprint()
```
slint, line 21: This can be removed I guess, since we merged to the above blueprint/resource.

invenio_communities/communities/service.py
```diff
@@ -10,10 +10,62 @@
...
 
 
 class CommunityService(RecordService):
     """community Service."""
 
-    pass
+    def search_user_communities(
+            self, identity, params=None, es_preference=None, **kwargs):
+        """Search for records matching the querystring."""
+        self.require_permission(identity, 'search')
```
slint, line 25: Shouldn't this be `search_user_communities` (which should then be set to `AuthenticatedUser()`)?

Then the `permission_action` for the resource should be just `read` (same way as it's done in drafts: https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/service.py#L49-L65)

slint (commented):
LGTM, one major issue on the permissions naming and usage in the service and resource method, and some other nits to be shelved.

invenio_communities/communities/service.py
```diff
@@ -10,10 +10,72 @@
...
+        )
+
+    def rename(self, id_, identity, data, revision_id=None):
+        """Rename a community."""
+        record = self.record_cls.pid.resolve(id_)
+
+        self.check_revision_id(record, revision_id)
+
+        # Permissions
+        self.require_permission(identity, "rename", record=record)
```
slint, line 60: Missed this one... We need a `can_rename` permission for this to work... Let's keep at a separate permission since in the future when we have multiple community members it will be only the community admin/owner that can rename or delete the community.

PR #242, opened 2021-04-21
communities: add settings/privileges tab

### LINKED ISSUE #228: views: privileges settings
image

zzacharo:
* Various UX fixes
closes #228 

# Screenshots
screencapture-127-0-0-1-5000-communities-mycom2-settings-2021-04-22-09_03_24

screencapture-127-0-0-1-5000-communities-r4ake-ac718-settings-privileges-2021-04-21-15_41_14

### DIFF
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -54,7 +54,7 @@ class CommunityMetadataSchema(Schema):
     ]
 
     title = SanitizedUnicode(required=True, validate=_not_blank(max=250))
-    description = SanitizedHTML(validate=_not_blank(max=2000))
+    description = SanitizedUnicode(validate=_not_blank(max=2000))
 
     curation_policy = SanitizedHTML(validate=_not_blank(max=2000))
     page = SanitizedHTML(validate=_not_blank(max=2000))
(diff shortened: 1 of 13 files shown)


invenio_communities/config.py
```diff
@@ -43,7 +43,7 @@
 ]
 
 # Overwite to change the default indexer for community records
-# communities_frontpageER_RECEIVER = None
+# communities_INDEXER_RECEIVER = None
```
slint, line 46: I think we can remove actually all the commented-out settings for now :)

invenio_communities/communities/schema.py
```diff
@@ -54,7 +54,7 @@ class CommunityMetadataSchema(Schema):
     ]
 
     title = SanitizedUnicode(required=True, validate=_not_blank(max=250))
-    description = SanitizedHTML(validate=_not_blank(max=2000))
+    description = SanitizedUnicode(validate=_not_blank(max=2000))
```
slint, line 57: The issue I see is that currently in Zenodo communities, `description` is a rich text field, and thus we would require to migrate the current `description` to some other field in order to preserve the original usage (cc @lnielsen). 

I can see that the purpose is to have an optional short tagline under the community's title, which differs from the original purpose... Let's discuss maybe what's the best design (or if we just take the toll when migrating)

slint (commented):
LGTM, let's discuss if we merge or create separate issues from other feedback as well.

PR #254, opened 2021-04-22
Fix tests and DataComponent usage

slint:


invenio_communities/communities/services/components.py
```diff
@@ -46,7 +47,8 @@ def _populate_access_and_validate(self, identity, data, record, **kwargs):
         if record is not None and "access" in data:
             # populate the record's access field with the data already
             # validated by marshmallow
-            record.update({"access": data.get("access")})
+            record.setdefault('access', {})
+            record['access'].update(data.get("access", {}))
```
slint, line 51: This is dangerous since we can override `owned_by`

PR #275, opened 2021-04-27
components: validates PID value on creation and renaming

### LINKED ISSUE #264: schema: `id` should not contain spaces
We need marshmallow validation for the `id` field:

- [ ] Only letters, numbers, and dashes/hyphens allowed

Glignos:
* Closes #264

### DIFF
--- a/invenio_communities/communities/services/components.py
+++ b/invenio_communities/communities/services/components.py
@@ -6,6 +6,7 @@
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
+import re
 from invenio_access.permissions import system_process
 from invenio_rdm_records.services.components import AccessComponent
 from invenio_records_resources.services.records.components import \
@@ -16,8 +17,18 @@ from marshmallow.exceptions import ValidationError
 class PIDComponent(ServiceComponent):
     """Service component for Community PIDs."""
 
+    @classmethod
+    def _validate(cls, pid_value):
+        """Checks the validity of the provided pid value."""
+        blop = re.compile('^[-\w]+$')
+        if not bool(blop.match(pid_value)):
+            raise ValidationError(
+                'The ID should contain only letters with numbers or dashes')
+
     def create(self, identity, record=None, data=None, **kwargs):
         """Create a Community PID from its metadata."""
+        data['id'] = data['id'].lower()
+        self._validate(data['id'])
         record['id'] = data['id']
         provider = record.__class__.pid.field._provider.create(record=record)
         setattr(record, 'pid', provider.pid)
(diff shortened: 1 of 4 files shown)


invenio_communities/communities/services/service.py
```diff
@@ -60,7 +60,7 @@ def search_user_communities(
             links_item_tpl=self.links_item_tpl,
         )
 
-    def rename(self, id_, identity, data, revision_id=None):
+    def rename(self, id_, identity, data, revision_id=None, raise_errors=True):
```
zzacharo, line 63: where is this `raise_errors` used?

invenio_communities/communities/services/components.py
```diff
@@ -6,18 +6,29 @@
...
 from invenio_access.permissions import system_process
 from invenio_rdm_records.services.components import AccessComponent
 from invenio_records_resources.services.records.components import \
     ServiceComponent
 from marshmallow.exceptions import ValidationError
 
 
+def validate_pid(pid_value):
+    """Checks the validity of the provided pid value."""
+    blop = re.compile('^[-\w]+$')
```
zzacharo, line 19: small comment on what this regex checking?

invenio_communities/communities/services/components.py
```diff
@@ -6,18 +6,29 @@
...
 
+import re
 from invenio_access.permissions import system_process
 from invenio_rdm_records.services.components import AccessComponent
 from invenio_records_resources.services.records.components import \
     ServiceComponent
 from marshmallow.exceptions import ValidationError
 
 
+def validate_pid(pid_value):
```
zzacharo, line 17: Maybe move this in the `PIDComponent` as `_validate`?

invenio_communities/communities/services/service.py
```diff
@@ -72,6 +72,14 @@ def rename(self, id_, identity, data, revision_id=None):
...
             raise ValidationError('Missing data for required field.', 'id')
 
+        data, errors = self.schema.load(
+            data,
+            context={"identity": identity},
+            raise_errors=raise_errors,     # if False, flow is continued with
+            schema_args={'partial': True}  # data only containing valid data,
+                                           # but errors are reported
+        )                                  # (as warnings)
+
```
slint, line 82: Isn't this problematic, since the `rename` data/payload only has the `id` field (and not e.g. the `metadata`, and `access` fields, which are required)?

invenio_communities/communities/services/service.py
```diff
@@ -72,6 +72,14 @@ def rename(self, id_, identity, data, revision_id=None):
         if 'id' not in data:
             raise ValidationError('Missing data for required field.', 'id')
 
+        data, errors = self.schema.load(
+            data,
+            context={"identity": identity},
+            raise_errors=raise_errors,     # if False, flow is continued with
```
zzacharo, line 78: Let's remove the `raise_errors` as we will always use the default which is True...

invenio_communities/communities/services/service.py
zzacharo: That's why I think the ` schema_args={'partial': True}` is passed so it doesn't require the whole schema...if I am not mistaken...

slint (approved):

PR #346, opened 2021-06-15
release: v2.3.2

ppanero:


CHANGES.rst
```diff
@@ -9,8 +9,12 @@
 Changes
 =======
 
+Version 2.3.2 (released 2021-06-15)
+-----------------------------------
+
+- Remove invenio dependencies to depend only on rdm-records.
 
-Version 2.3.1 (released 2021-06-10)
+Version 2.3.1 (removed)
```
ppanero, line 17: @lnielsen is this correct? or how do we state that it was taken down from PyPi?

ntarocco: I think it is ok like this, so we keep track of it...

ntarocco (approved):

PR #361, opened 2021-11-10
records: initial community/record data layer

lnielsen:
addresses #355


invenio_communities/records/records/systemfields/communities/manager.py
```diff
@@ -0,0 +1,205 @@
...
+                'Cannot set community as the default. '
+                'The record has not been added to the community.'
+            )
+        self._default_id = id_
+
+    @default.deleter
+    def default(self):
+        self._default_id = None
+
+    # Persist relationships in record (denormalize the M2M table).
```
lnielsen, line 186: In the end I went for denormalising the entire structure into the parent record so looks like this (originally I was only going to add ``"default"``:

```
{
    "default": "<uuid>",
    "ids": ["<uuid>", "<uuid>", ...],
}
```

invenio_communities/records/records/systemfields/communities/field.py
```diff
@@ -0,0 +1,56 @@
...
+        self.set_dictkey(record, manager.to_dict())
+
+    #
+    # Helpers
+    #
+    def obj(self, record):
+        """Get or crate the communities manager."""
+        # Check cache
+        obj = self._get_cache(record)
+        if obj is not None:
```
lnielsen, line 41: It took me an hour to track down a bug related to this line 🥸 A classic one of ``if obj`` vs ``if obj is None`` - even though we got an object from the cache, it sometimes evaluated to false because we implemented ``__len__()``

max-moser: at some point we should probably check the code-base for such potential bugs :detective: 

one occasion where they're actually warranted is when checking `LocalProxy`'d objects – in this case, the `… is None` check will never evaluate to `True`, even if the object behind the proxy is set to `None` (after all, the checked reference is still the proxy)
in such cases, the `not …` will still work however, since i think it does an implicity `bool(…)` cast and thus resolves the object behind the proxy
example: https://github.com/inveniosoftware/invenio-userprofiles/blob/master/invenio_userprofiles/api.py#L30 where the `current_user` proxy is used

invenio_communities/records/records/systemfields/communities/manager.py
```diff
@@ -0,0 +1,205 @@
...
+
+    @property
+    def ids(self):
+        """Get communities ids."""
+        return sorted(self._communities_ids)
+
+    @property
+    def default(self):
+        """Get the default community."""
+        if self._default_id:
```
max-moser, line 163: could it happen here too, that the `_default_id` is something falsy but different from `None`?
e.g. `id == 0` (unlikely b/c it's usually UUIDs, but still)?

max-moser (approved):
LGTM :+1:

invenio_communities/records/records/systemfields/communities/manager.py
lnielsen: Yep, better safe than sorry - don't want to spend another hour debugging :D

PR #380, opened 2022-01-31
membership: invite user to community

fenekku:


tests/memberships/test_membership_services.py
```diff
@@ -0,0 +1,176 @@
...
+    return {
+        "type": "community-member-invitation",
+        "receiver": {"user": another_user.id},
+        "topic": {'community': community.data['uuid']}
+    }
+
+
+# Tests
+
+def test_invite_user_flow(
```
fenekku, line 116: Start the review here: high-level workflow.

tests/memberships/test_membership_services.py
```diff
@@ -0,0 +1,176 @@
...
+        another_identity, community, community_owner_identity,
+        community_service, invitation_creation_input_data,
+        requests_service):
+    community_id = community.data['uuid']
+
+    # Invite
+    invitation = community_service.invitations.create(
```
fenekku, line 123: Something to note (maybe a smell), we are using 
- the invitations service to create an invitation
- the requests service to accept it
- the members service to check the final membership

It makes sense... but maybe we are just leaking the underlying abstractions...

tests/memberships/test_membership_services.py
```diff
@@ -0,0 +1,176 @@
...
+    assert invitation_dict['is_open'] is False
+    assert invitation_dict['is_closed'] is True
+
+    # Get membership
+    member = community_service.members.read(
+        community_owner_identity,
+        {
+            "community": community_id,
+            "user": str(another_identity.id),
+        }
```
fenekku, line 150: 1- In a future PR I will clarify the distinction between community id and community uuid. Here community_id means community uuid. I think at the data layer level we want uuid used and at the resource level, the id. This means the service would have to convert between one and the other. Do we have mechanisms to do so?

2- Because I see the endpoint as: `/api/communities/<community_id>/members/<user|group>/<entity_id>`, the id for the `read` method is a composite id. This hasn't really been the case for services in the past I think. Thoughts?

invenio_communities/members/services/permissions.py
```diff
@@ -0,0 +1,182 @@
+# -*- coding: utf-8 -*-
```
fenekku, line 1: Ignore for now until permissions are addressed in future PR.

invenio_communities/members/resources/resource.py
```diff
@@ -0,0 +1,142 @@
+# -*- coding: utf-8 -*-
```
fenekku, line 1: Ignore for now until resources are addressed in future PR.

invenio_communities/members/resources/config.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
fenekku, line 1: Ignore for now until resources are addressed in future PR.

invenio_communities/members/resources/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
```
fenekku, line 1: Ignore for now.

invenio_communities/members/records/models.py
```diff
@@ -0,0 +1,36 @@
...
+        db.ForeignKey(CommunityMetadata.id, ondelete="CASCADE"),
+        primary_key=True
+    )
+    user_id = db.Column(
+        db.Integer(),
+        db.ForeignKey(User.id, ondelete="CASCADE"),
+        primary_key=True,
+        nullable=True)
+    # TODO: add group
+    # group_id = db.Column(db.Integer(), primary_key=True)
```
fenekku, line 36: Migrations will only be made and committed at the end of the sprint.

invenio_communities/members/records/jsonschemas/community_members/member-v1.0.0.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "id": "local://community_members/member-v1.0.0.json",
+  "additionalProperties": false,
+  "title": "Invenio Community Member Schema v1.0.0",
+  "type": "object",
+  "properties": {
+    "$schema": {
+      "$ref": "local://definitions-v1.0.0.json#/$schema"
+    }
```
fenekku, line 10: More in next PRs.

invenio_communities/members/records/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # created_by = EntityReferenceField("created_by", check_referenced)
+    # """Who created the event."""
+
+    @classmethod
+    def get_record(cls, id_, with_deleted=True):
+        """Return Member."""
+
+        with db.session.no_autoflush:
+            query = cls.model_cls.query.filter_by(**id_)
```
fenekku, line 71: Strange, but it is what it is.

invenio_communities/invitations/services/service.py
```diff
@@ -0,0 +1,75 @@
...
+            community_dict).resolve()
+
+        type_ = current_registry.lookup(data.pop('type', None))
+
+        entity = ResolverRegistry.resolve_entity_proxy(
+            entity_dict).resolve()
+
+        # Check if entity is already a member of community
+        try:
+            current_communities.service.members.read(
```
fenekku, line 51: Perhaps a little clumsy: using the members service in the invitations service to check against prior membership. We will also need to check against prior invitation. (Putting it in a component would be hiding the weirdness rather than fixing it...)

invenio_communities/invitations/services/request_types.py
```diff
@@ -0,0 +1,40 @@
...
+        "open": RequestState.OPEN,
+        "cancelled": RequestState.CLOSED,
+        "declined": RequestState.CLOSED,
+        "accepted": RequestState.CLOSED,
+        "expired": RequestState.CLOSED,
+    }
+    default_status = "open"
+    available_actions = {
+        "accept": AcceptAction,
+        # "cancel": CancelAction,
```
fenekku, line 31: Todo in next PRs.

invenio_communities/communities/services/service.py
```diff
@@ -20,16 +20,30 @@
...
 
-    def __init__(self, config, files_service=None):
+    def __init__(
+            self, config, files_service=None, invitations_service=None,
+            members_service=None):
         """Constructor for CommunityService."""
         super().__init__(config)
         self._files = files_service
+        self._invitations = invitations_service
+        self._members = members_service
```
fenekku, line 30: Placing them all under the communities service... The members and invitations services are pretty coupled. Maybe they are not super coupled to the communities one. The programmatic API is nicer looking this way :shrug:.

invenio_communities/members/records/models.py
```diff
@@ -0,0 +1,36 @@
...
+    community_id = db.Column(
+        UUIDType,
+        db.ForeignKey(CommunityMetadata.id, ondelete="CASCADE"),
+        primary_key=True
+    )
+    user_id = db.Column(
+        db.Integer(),
+        db.ForeignKey(User.id, ondelete="CASCADE"),
+        primary_key=True,
+        nullable=True)
```
fenekku, line 34: I've seen the nullable and primary key threads (e.g. https://github.com/inveniosoftware/invenio-access/issues/188). We'll probably need to add partial indexes.

fenekku (commented):
Feedback most welcome. We can discuss at standup.

max-moser:
unfortunately, i don't know about the alembic error...
i just re-ran the checks on master, and they fail too (with a slightly different error, though), which sounds a bit like a files-related module to me:
`AssertionError: assert not [('remove_index', Index('ix_uq_partial_files_object_is_head', Column('bucket_id', UUID(), table=<files_object>, nullable=False), Column('key', TEXT(), table=<files_object>, nullable=False)))]`

invenio_communities/invitations/services/actions.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Minor: can we follow same pattern as already used here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/requests/community_submission.py - a request all in one file

invenio_communities/invitations/services/request_types.py
```diff
@@ -0,0 +1,40 @@
...
+from invenio_requests.customizations import RequestState, BaseRequestType
+
+from .actions import AcceptAction
+
+
+class CommunityMemberInvitation(BaseRequestType):
+    """Community member invitation request type."""
+
+    type_id = 'community-member-invitation'
+    name = _('Community Member Invitation')
```
lnielsen, line 20: This will be displayed in the request detail page so go with either 

- Invitation
- Community invitation
- Member invitation

Also only [first letter in first word in capital](https://inveniordm.docs.cern.ch/maintenance/styleguide/#captilization)

Screenshot 2022-02-02 at 11 05 00

invenio_communities/invitations/services/actions.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def execute(self, identity, uow):
+        """Accept entity into community."""
+        # community = self.request.topic.resolve()
+        # entity = self.request.receiver.resolve()
+
+        member_data = {
+            **self.request.receiver.reference_dict,
+            **self.request.topic.reference_dict,
+            # TODO: add role
```
lnielsen, line 34: Also visibility (public/hidden) - default to hidden

invenio_communities/members/records/models.py
```diff
@@ -0,0 +1,36 @@
...
+
+from invenio_db import db
+from invenio_accounts.models import User
+from invenio_records.models import RecordMetadataBase
+from sqlalchemy_utils.types import UUIDType
+
+from ...communities.records.models import CommunityMetadata
+
+
+class MemberModel(db.Model, RecordMetadataBase):
```
lnielsen, line 18: I don't think we should use a record with json column as table. I think it should look like:


- ``user_id`` - Foreign key to the user
- ``group_id`` - Foreign key to the role
- ``community_id`` - Foreign key to the community
- ``role`` - Role identifier (str) e.g. ``owner``, ``manager``
- ``visibility`` - Boolean (true=public, false=hidden)
- ``created`` - Creation timestamp (via ``Timestamp`` model from SQLAlchemy-Utils)
- ``updated`` - Modification timestamp. (via ``Timestamp`` model from SQLAlchemy-Utils)

We need indices on ``user_id``, ``group_id`` (select all memberships for a user), and ``community_id`` (select all memberships for a community.

Example of a query we have to do:

```sql
SELECT DISTINCT 
    community_id, role
FROM 
    community_members 
WHERE 
    user_id=<user_id> OR 
    group_id IN [<group_id>, <group_id>, ...,]
```

Also, possibly we should see if we could add a check constraint to ensure only group or user is set and not both.

invenio_communities/members/records/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+from invenio_db import db
+from invenio_records.systemfields import ConstantField, DictField, ModelField
+from invenio_records_resources.records.api import Record
+from invenio_records_resources.records.systemfields import IndexField
+
+from .models import MemberModel
+
+
+class Member(Record):
```
lnielsen, line 16: See notes on the SQLAlchemy model which will impact this API.

invenio_communities/members/resources/resource.py
lnielsen: Comment: To merge just make sure these files are cleaned out if not done

invenio_communities/members/services/service.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+        # It's the components that save the actual data in the record.
+        member_record = self.record_cls.create(
+            {},
+            **self._to_id(data),
+        )
+
+        # Run components
+        self.run_components(
```
lnielsen, line 44: Minor: I not sure we need components. It's mainly for when a service has multiple independent concerns, and the members service has a single responsibility. Thus, I'd rather keep it out to simplify and see if we really need it. See https://inveniordm.docs.cern.ch/develop/topics/service/#service-components

lnielsen (commented):
I think we should likely create a new branch ``members`` in the repositories that we can merge to. As multiple teams work on the repository, we'll need to have the code clean when merging.

invenio_communities/members/records/models.py
fenekku: Will do migrations now so that invenio-rdm-records master doesn't break.

invenio_communities/communities/services/service.py
fenekku: Agreed. I am waiting to see if the abstraction will reveal itself as we go.

invenio_communities/invitations/services/actions.py
fenekku: sure

invenio_communities/invitations/services/request_types.py
fenekku: What will we call it when we invite records to be part of the community? I was thinking a parallel naming structure:

- Community member invitation
- Community record invitation

I've also been prefacing the names with "Community", so that they are namespaced. Member / Record... is very generic and would require namespacing by other modules down the road anyway.

invenio_communities/members/records/models.py
fenekku: I see, might as well not inherit `RecordMetadataBase` :+1:

invenio_communities/members/services/service.py
fenekku: sure

invenio_communities/members/records/models.py
fenekku: Unfortunately, removing RecordMetadataBase has many unintended consequences: invenio-records/invenio_records/api.py assumes that the underlying model has all the fields of a RecordMetadataBase. (uses keyword arguments in `create`). Overriding the api.py Record is tricky for me because I don't know if most of it should be kept / roles of hte different signals/pre/post extensions. 

I will keep the inheritance but disable the field if I can.

fenekku:
The same 

```
assert not ext.alembic.compare_metadata()
E       AssertionError: assert not [('remove_index', Index('ix_uq_partial_files_object_is_head', Column('bucket_id', UUID(), table=<files_object>, nullable=False), Column('key', TEXT(), table=<files_object>, nullable=False)))]
```

error that is affecting the master branch, invenio-rdm-records PRs and basically every repository using invenio-files-rest is affecting this PR too. All the tests pass locally, but I will feel better when this is solved prior to merging. Some of the remarks I couldn't quite address anyway, so it might be worth taking another look.

lnielsen (approved):

lnielsen:
Not further points than the ones already discussed.

PR #388, opened 2022-02-03
382 membership id

### LINKED ISSUE #382: Add members primary key id.

fenekku:
- memberships: add id as primary key
- closes #382

### DIFF
--- a/invenio_communities/members/records/api.py
+++ b/invenio_communities/members/records/api.py
@@ -6,7 +6,7 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 from invenio_db import db
-from invenio_records.systemfields import ConstantField, ModelField
+from invenio_records.systemfields import ModelField
 from invenio_records_resources.records.api import Record
 from invenio_records_resources.records.systemfields import IndexField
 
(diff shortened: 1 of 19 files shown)


invenio_communities/members/records/api.py
```diff
@@ -0,0 +1,49 @@
...
+    """A Request Event."""
+
+    model_cls = MemberModel
+
+    # Systemfields
+
+    metadata = None
+
+    index = IndexField(
+        "community_members-members-v1.0.0", search_alias="community_members"
```
lnielsen, line 26:
```suggestion
        "communitymembers-members-v1.0.0", search_alias="communitymembers"
```

invenio_communities/members/records/api.py
```diff
@@ -0,0 +1,49 @@
...
+
+from invenio_db import db
+from invenio_records.systemfields import ConstantField, ModelField
+from invenio_records_resources.records.api import Record
+from invenio_records_resources.records.systemfields import IndexField
+
+from .models import MemberModel
+
+
+class Member(Record):
```
lnielsen, line 16: Same comment here as for invitations in that it should not be a record but simply a sqlalchemy model (similar to record/community tables. I

lnielsen (approved):

invenio_communities/members/records/api.py
fenekku: After implementing permissions (https://github.com/inveniosoftware/invenio-communities/pull/394), having Member be a Record actually seems like a good thing. Let's discuss this again perhaps?

PR #393, opened 2022-02-08
api: refactor facets definition

zzacharo:
# Screenshots
Screenshot 2022-02-09 at 16 13 55

### DIFF
--- a/invenio_communities/communities/services/config.py
+++ b/invenio_communities/communities/services/config.py
@@ -22,6 +22,7 @@ from invenio_records_resources.services.records.links import RecordLink, \
     pagination_links
 
 from invenio_communities.communities.records.api import Community
+from invenio_communities.communities.services import facets
 
 from ..schema import CommunitySchema
 from .components import CommunityAccessComponent, PIDComponent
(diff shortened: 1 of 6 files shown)


invenio_communities/communities/services/config.py
```diff
@@ -32,18 +33,7 @@ class SearchOptions(SearchOptionsBase):
...
-            field='metadata.type',
-            label=_('Type'),
-            value_labels={
-                'organization': _('Organization'),
-                'event': _('Event'),
-                'topic': _('Topic'),
-                'project': _('Project'),
-            }
-        ),
-        'domain': TermsFacet(field='metadata.domains'),
```
jrcastro2, line 45: In the mockups there are 2 facets and you only keep one, was this removed by mistake or on purpose?

jrcastro2 (commented):
LGTM! Just one small comment

invenio_communities/communities/services/config.py
zzacharo: The domain was already defined as a facet but the field is not implemented at all. In the mockups, we have facets for 
- Visibility
- Ownership

I implemented the first one but the second one I think is not there yet. I updated the screenshot of how it looks like now

jrcastro2 (approved):

PR #399, opened 2022-02-11
config: make service configs configurable

mirekys:
This enables users of the module to override default configs of the provided services.

One could implement & provide their own `ServiceConfig` class inheriting from `CommunityServiceConfig`,
and e.g. customize PermissionPolicies, record links, add fields to Community records through
extending record schema, etc.

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -9,6 +9,9 @@
 """Default configuration."""
 
 from invenio_communities.communities.services import facets
+from invenio_communities.communities import CommunityServiceConfig, CommunityFileServiceConfig
+from invenio_communities.invitations import InvitationServiceConfig
+from invenio_communities.members import MemberServiceConfig
 
 COMMUNITIES_MEMBERSHIP_REQUESTS_CONFIRMLINK_EXPIRES_IN = 1000000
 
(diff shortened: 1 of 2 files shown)


lnielsen (commented):
First of all, thanks for the contribution and apologies I didn't see this earlier.

We're trying to avoid the pattern used here as it opens up too large an API that we're likely not going to be able to keep stable, and often the configs might be interconnected making it hard for users of InvenioRDM to change it. 

Please see the example here: 
https://inveniordm.docs.cern.ch/develop/topics/service/#instantiating-a-service 

```python
service = ClickService(ClickServiceConfig.build(current_app))
service.click(system_identity)
```

You'll find the pattern implemented in Invenio-RDM-Records - e.g. https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L99-L102

Which specific items do you want to configure? Just permissions?

mesemus:
Hi Lars,

in short:

thanks for the reply, I understand why you make this design decision even although 
it brings a couple of problems to our repositories :) Also a lot of stuff has been changed 
to be configurable via FormConfig, so for majority of users your approach will work without
problems.

This PR can be closed as we have a workaround for this - we keep our own set 
of patches to invenio and can have this extra functionality in them.

More context:

In our repositories, we have records of multiple types and we would like to reuse 
most of the functionality that is written around RDM records. We can not 
reuse the RDM records themselves as our metadata schemas are quite different.

By extending invenio-rdm-records's service we have created a "rdm compatibility" 
layer that in most cases pretends to the rest of Invenio that our records are RDMish 
(works almost everywhere where invenio uses service layer to access records and 
is not peckish about the format of the metadata section), but there are places where 
this does not work (either RDMRecord is imported directly and used without service,
or some parts of Invenio depend on UI implementation of RDM records). For these, 
we need to overwrite them and provide custom implementation - and here replaceable 
configs & implementation classes allow us to do it cleanly.

But as I said above, our use-case is not the common one and we are happy to keep
this in our own patches, so let's close this PR.

> First of all, thanks for the contribution and apologies I didn't see this earlier.
> 
> We're trying to avoid the pattern used here as it opens up too large an API that we're likely not going to be able to keep stable, and often the configs might be interconnected making it hard for users of InvenioRDM to change it.
> 
> Please see the example here: https://inveniordm.docs.cern.ch/develop/topics/service/#instantiating-a-service
> 
> ```python
> service = ClickService(ClickServiceConfig.build(current_app))
> service.click(system_identity)
> ```
> 
> You'll find the pattern implemented in Invenio-RDM-Records - e.g. https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L99-L102
> 
> Which specific items do you want to configure? Just permissions?

PR #417, opened 2022-02-15
tests: put back alembic test

ntarocco:


fenekku:
Fixed in https://github.com/inveniosoftware/invenio-communities/pull/418. Thanks!

PR #418, opened 2022-02-15
tests: fix alembic tests

fenekku:


invenio_communities/members/records/models.py
```diff
@@ -27,6 +27,7 @@ class MemberModel(db.Model, RecordMetadataBase):
     community_id = db.Column(
         UUIDType,
         db.ForeignKey(CommunityMetadata.id, ondelete="CASCADE"),
+        nullable=False
```
fenekku, line 30: It was this little bugger...

fenekku:
Merging this to unblock invenio-rdm-records.

invenio_communities/members/records/models.py
ntarocco: I was so focused to check the Alembic recipes, that I didn't have a look to the `models`... good job and thanks for fixing it! :)

PR #419, opened 2022-02-16
styles: minor fixes

jrcastro2:


invenio_communities/assets/semantic-ui/less/invenio_communities/theme.less
```diff
@@ -113,26 +111,27 @@
...
-    margin-top: -8px;
+  label.helptext {
     margin-bottom: 15px;
     display: block;
     color: #737373;
     font-weight: normal;
     font-size: 0.9em;
   }
 
+  .ui.header:first-child {
```
kpsherva, line 125: this will affect all the semantic ui headers in the community module or any place which uses its css assets, is this the goal?

jrcastro2: It was adding negative margin I guess it's needed by the cmp itself so will make it specific :+1: Thanks!

invenio_communities/assets/semantic-ui/less/invenio_communities/theme.less
```diff
@@ -143,14 +139,22 @@
...
     height: 173px;
     overflow: hidden;
   }
+}
 
+.community-records-search {
+  .ui.container {
+    .ui.relaxed.grid {
+      .result-options {
+        margin-top: 0em;
```
kpsherva, line 156: nit: 0 does not need a unit

invenio_communities/assets/semantic-ui/less/invenio_communities/theme.less
```diff
@@ -143,14 +139,22 @@
...
     max-height: 173px;
     max-width: 209px;
     height: 173px;
     overflow: hidden;
   }
+}
 
+.community-records-search {
+  .ui.container {
+    .ui.relaxed.grid {
```
kpsherva, line 154: I am still not sure about this, sorry :) there is too much parent child dependencies in this rule. Fe. if someone changes the `.relaxed` class, the rule will break immediately. if you have time we could maybe see it together

invenio_communities/assets/semantic-ui/less/invenio_communities/theme.less
```diff
@@ -7,9 +7,8 @@
  */
 
 .community-header {
-  background: #F8F8F9;
+  background: #f8f8f9;
```
Pineirin, line 10: Why aren't we defining here at least the colors as a variable? It would make it easier to maintain, but maybe there is a reason for not doing it.

zzacharo: Having a variable is necessary I agree :)

jrcastro2: Agreed with @kpsherva to fix this in next issues (https://github.com/inveniosoftware/invenio-theme/issues/265)

Pineirin: Ok, apart from that LGTM!

Pineirin (approved):

kpsherva (approved):

Pineirin (approved):
LGTM!

Issue #329, opened 2021-04-30
Missing community pic in privileges page


fenekku:
Package version (if known): April release (invenio-communities 2.2.5)

## Describe the bug
When adding a community logo the profile page displays it nicely, but the privilege page doesn't.

Profile page (and elsewhere too)

![community-profile-has-pic](https://user-images.githubusercontent.com/1932651/116757440-ad8fad00-a9d3-11eb-9285-62c4d2e38228.png)

Privileges page

![community-privilege-doesnt-have-pic](https://user-images.githubusercontent.com/1932651/116757472-c39d6d80-a9d3-11eb-9c07-88aa4ce51ba9.png)

## Steps to Reproduce
1. Go to '/communities/new'
2. Fill it out and create the community
3. Upload picture on community profile page
4. Go to /privileges page

## Expected behavior
Profile pic is there too.

I know this is probably due to /privileges being static and the rest of the thumbnail occurrences being populated via API.

PR #429, opened 2022-02-22
privilegesPage: fix missing community logo

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-communities/issues/329

image

### DIFF
--- a/invenio_communities/views/communities.py
+++ b/invenio_communities/views/communities.py
@@ -101,9 +101,10 @@ def communities_settings(community=None, logo=None, pid_value=None):
         active_menu_tab="settings"
     )
 
-@pass_community
 @require_community_owner
-def communities_settings_privileges(community=None, pid_value=None):
+@pass_community
+@pass_community_logo
+def communities_settings_privileges(community=None, logo=None, pid_value=None):
     """Community settings/privileges page."""
     return render_template(
         "invenio_communities/details/settings/privileges.html",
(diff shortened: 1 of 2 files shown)


invenio_communities/views/communities.py
```diff
@@ -138,5 +139,6 @@ def communities_settings_privileges(community=None, pid_value=None):
         # Pass permissions so we can disable partially UI components
         # e.g Settings tab
         permissions=community.has_permissions_to(['update']),
-        active_menu_tab="settings"
+        active_menu_tab="settings",
+        logo=logo.to_dict() if logo else None
```
NRodriguezcuellar, line 143: Interesting, so the logo is an object, can it be serialized by the view?

Pineirin (approved):
LGTM!

invenio_communities/views/communities.py
```diff
@@ -103,7 +103,8 @@ def communities_settings(community=None, logo=None, pid_value=None):
 
 @pass_community
 @require_community_owner
-def communities_settings_privileges(community=None, pid_value=None):
+@pass_community_logo
+def communities_settings_privileges(community=None, logo=None, pid_value=None):
```
ntarocco, line 107: Minor: can you move it up, as for the other view func, so that the `require_community_owner` will be evaluated first (if I am not mistaken...)? I mean, we should check permissions first.

Can you please please change the 3 decorators implementation and have:
```
pid_value = kwargs['pid_value']
```
instead of
```
pid_value = kwargs.get('pid_value')
```
?

invenio_communities/views/communities.py
jrcastro2: The view is supposed to only read the object, that for we have [this function](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/services/service.py#L114-L127) in the service which is called int the `pass_community_logo` decorator and with the to_dict() we get the dict representation of the logo object to be used in the view. Hope this answers your question.

kpsherva (approved):

PR #440, opened 2022-03-01
search: fix header

### LINKED ISSUE #439: Communities: wrong header spacing
Screenshot 2022-03-01 at 18 03 52

ntarocco:
* closes #439

Not very nice, better than current, facets are missing at the moment...

Screenshot 2022-03-01 at 18 53 07


ntarocco:
@kpsherva we need really to rework all the gray headers: there is code duplicated everywhere, with different implementations. there should be only one top banner and one header defined somewhere.

lnielsen (approved):

PR #445, opened 2022-03-07
tests: highlight problem with multiple communities creation

fenekku:


tests/communities/test_service.py
```diff
@@ -0,0 +1,140 @@
...
+    """Problematic case."""
+
+    communities = [
+        invalid_community_input,
+        valid_community_input_1,
+        valid_community_input_2
+    ]
+
+    # Expect the same results as above but it doesn't happen!
+    assert 1 == number_errors(communities)
```
fenekku, line 140: number errors is 2. See https://github.com/inveniosoftware/invenio-communities/issues/446 for relevant stack trace

PR #523, opened 2022-04-08
invitation: refresh index after inviting

ntarocco:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/320


invenio_communities/members/services/service.py
```diff
@@ -114,6 +117,13 @@ def invite(self, identity, community_id, data, uow=None):
             uow
         )
 
+        # ensure index is refreshed to search for newly added members
+        # and invitations
+        uow.register(
```
lnielsen, line 122: I think with the changes from #524 you only need to refresh the members index.

lnielsen (approved):

PR #545, opened 2022-04-22
search: return communities a user is member of via search_user_communities

### LINKED ISSUE #503: Backend: Communities: search_user_communites() should only return communities for the given user
The ``CommunityService.search_user_communities()`` should only return communities that the given identity is a member of.

Essentially, we need to apply a query filter:

```
extra_filter=Q("terms", id=[com_uuid1, com_uuid2, ...])
```

Likely the easiest is to use the "CommunitMembers" generator to extra the list of communities from the identity, so it becomes something like:

```python
search_result = self._search(
    #...
    extra_filter=CommunityMembers().query_filter(identity=identity)
)
```

A tests should be added to show that it works.

Out of scope: Right now we don't have support for saying that an identity have access to **all** communities (e.g. super admin).

fenekku:
closes #503

### DIFF
--- a/invenio_communities/generators.py
+++ b/invenio_communities/generators.py
@@ -176,7 +176,7 @@ class CommunityRoles(Generator):
         # Gives access to all community members.
         return Q(
             "terms",
-            **{"id": self.communities(identity)}
+            **{"_id": self.communities(identity)}
         )
 
 
@@ -188,7 +188,8 @@ class CommunityMembers(CommunityRoles):
 
     def communities(self, identity):
         return [
-            n.value for n in identity.provides if n.method == 'community']
+            n.value for n in identity.provides if n.method == 'community'
+        ]
 
 
 class CommunityManagers(CommunityRoles):
--- a/tests/communities/test_services.py
+++ b/tests/communities/test_services.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2022 Graz University of Technology.
+# Copyright (C) 2022 Northwestern University.
 #
 # Invenio-Communities is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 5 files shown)


invenio_communities/generators.py
```diff
@@ -176,7 +176,7 @@ def query_filter(self, identity=None, **kwargs):
         # Gives access to all community members.
         return Q(
             "terms",
-            **{"id": self.communities(identity)}
+            **{"_id": self.communities(identity)}
```
lnielsen, line 179: Question: Is this a fix to a bug you've found?

lnielsen (commented):
Thanks a lot! Just smaller comments from me.

fenekku:
Didn't have time until today to look into this. Looking at this now.

invenio_communities/generators.py
fenekku: Yes I suppose. `"id"` for a community is still the slug, but the CommunityRoleNeed stores the uuid, so the permission search didn't work without this change. What I've noticed was that a `uuid` (literally) field was also present in the document during tests, but I couldn't find its matching official mapping, so I went for `_id` which is guaranteed to be there. It is a little strange given that `_id` is the ES document id, but again I didn't see another official field where the uuid is kept.

fenekku:
changes made.

tests/communities/test_services.py
```diff
@@ -206,8 +212,72 @@ def test_update_featured(community_service, comm):
 
 def test_cleanup_pre_search(db, es_clear):
     """Cleanup database and elasticsearch for following tests.
-    
+
     Tests before do not depend on a clean db or es. Cleanup is also an
     expensive task so te tests will run faster without them.
     """
     pass
+
```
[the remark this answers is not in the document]
fenekku: done

fenekku:
How can we improve this situation where a PR is made, but left hanging for weeks, so eventually other releases break its tests? 

Perhaps there is a systematic change we can do to help... :thinking:

fenekku:
rebased

lnielsen:
Thanks for updating. The rate of change has been so high on some of the repositories and with big far reaching changes, that even PRs made the same day would break. Definitely we can improve the situation, and we will discuss how, but it's also not unavoidable.

Issue #546, opened 2022-04-22
Remove hard-coded search app configuration from html.


Pineirin:
In the file https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/search.html the configuration of the searchapp is hard-coded, instead of being configured in the searchapp.py file. This should be fixed.

PR #547, opened 2022-04-25
Communities search: define search configuration in searchapp.py

Pineirin:
closes: inveniosoftware/invenio-communities#546


invenio_communities/config.py
```diff
@@ -109,6 +115,11 @@
 ]
 """Community roles."""
 
+COMMUNITIES_SEARCH = {
+    'facets': ['type', 'domain'],
```
jrcastro2, line 119: COMMUNITIES_FACETS also contains visibility (see [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/config.py#L44)), I guess it should be added here or removed from the facets.

invenio_communities/config.py
```diff
@@ -109,6 +115,11 @@
...
+    'sort': ['bestmatch', 'newest', 'oldest', 'version'],
```
jrcastro2, line 120: I think communities don't have version, or do they?

Pineirin: Right, they don't.

invenio_communities/templates/semantic-ui/invenio_communities/search.html
```diff
@@ -22,82 +22,7 @@
...
-      "aggName": "type",
-      "field": "metadata.type",
-      "title": "{{_("Type")}}"
-    },
-    {
-      "aggName": "domain",
-      "field": "metadata.domains",
-      "title": "{{_("Domain")}}"
-    }
-  ],
```
Pineirin, line 37: Facets where defined in configuration to mimic the hard-coded ones.

invenio_communities/config.py
Pineirin: I'm a bit unsure about this, I tried to keep the same facets as the one that were previously hard-coded: https://github.com/inveniosoftware/invenio-communities/pull/547/files#r857564267

jrcastro2: There might be some things in the code that are not the best it could be and hardcoded things are usually a red flag for this, so if we can improve something we should take the opportunity :wink: 
As far as I understand and if I am not mistaken if it's declared in the COMMUNITIES_FACETS this means that it will generate the aggregation for the field (in this case `visibility`) and in the COMMUNITIES_SEARCH we establish what we want  to display in the search, if we do the aggregation I think we should use it, otherwise we should not do it.

invenio_communities/communities/services/facets.py
```diff
@@ -48,4 +48,15 @@
         'hidden': _('Hidden'),
         'restricted': _('Restricted'),
     }
-)
\ No newline at end of file
+)
+
+domain = TermsFacet(
```
ntarocco, line 53: to be removed

invenio_communities/config.py
```diff
@@ -46,7 +46,13 @@
         'ui': {
             'field': 'visibility',
         }
-    }
+    },
+    'domain': {
```
ntarocco, line 50: to be removed

invenio_communities/config.py
ntarocco: As discussed, to keep only `type`

ntarocco (approved):

ntarocco (approved):

zzacharo (approved):

Issue #509, opened 2022-04-06
Backend: Communities: Sort members by name instead of newest


lnielsen:
- [x] `member_service.search_public()/search()` should be default sort by name instead of newest (and newest should not be available on public).
- [x] `member_service.search_public()` should be default sort on best match and member name

PR #574, opened 2022-04-29
members: add sort by name

ppanero:
- closes https://github.com/inveniosoftware/invenio-communities/issues/509

Missing more in depth tests.

### DIFF
--- a/tests/members/test_members_resource.py
+++ b/tests/members/test_members_resource.py
@@ -10,7 +10,9 @@
 
 import pytest
 
+from invenio_access.permissions import system_identity
 from invenio_requests.records.api import RequestEvent
+
 
 #
 # Fixtures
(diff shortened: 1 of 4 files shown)


tests/members/test_members_resource.py
```diff
@@ -216,30 +247,36 @@ def test_search(
     assert hit['permissions']['can_update_visible'] is True
     assert hit['visible'] is True
 
-    hit = data['hits']['hits'][1]
+    hit = data['hits']['hits'][1]  # should be last, test sorting
```
ppanero, line 250: To test the `name` sorting I added a user that would be in the middle of the two previous ones (S < N, and then the third one has no name and gets "Untilted" when serialized).

However, this gives me the feeling of being too ad-hoc. On the other hand, I'm not sure if it's worth making a full test just for sorting.

ntarocco (approved):

lnielsen (approved):

zzacharo (approved):

PR #580, opened 2022-05-03
global: enhance resolver and add uuid to ES mappings

ntarocco:
* adds the UUID field to the ES mapping as keyword
* enhances the community resolver to inject the service and pick the
  fields to be expanded when resolved
* removes dependency invenio-users-resources, added to invenio-requests
* overrides the default service read_many to search by uuid

requires https://github.com/inveniosoftware/invenio-requests/pull/202


ntarocco:
Tests are failing because of https://github.com/inveniosoftware/pytest-invenio/commit/029ebc8038774ab5301ab3936dd5aa4650d29941, to be checked after merge to unblock the build of other modules.

PR #531, opened 2022-04-13
Communities relations

ctrl-alt-soneca:
- requires https://github.com/inveniosoftware/invenio-vocabularies/pull/187
- closes https://github.com/inveniosoftware/invenio-communities/issues/500
- closes https://github.com/inveniosoftware/invenio-communities/issues/501

:heart: Thank you for your contribution!

### Description

Please describe briefly your pull request.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x]I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

PR #582, opened 2022-05-03
communities: added funding to community settings.

alejandromumo:
- requires https://github.com/inveniosoftware/invenio-communities/pull/531 Otherwise the `funding` relation is not going to be denormalised and the UI won't render funders/awards properly.
- requires https://github.com/inveniosoftware/react-invenio-deposit/pull/465 being released. Otherwise the component will not be found. 
- closes https://github.com/inveniosoftware/invenio-communities/issues/587

this PR adds `FundingField` to community settings.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
@@ -29,6 +29,7 @@ import Dropzone from "react-dropzone";
 
 import { DeleteButton } from "./DeleteButton";
 import { RenameCommunityButton } from "./RenameCommunityButton";
+import { FundingField } from "react-invenio-deposit";
 
 import {
   Divider,
@@ -253,6 +254,7 @@ class CommunityProfileForm extends Component {
         type: {},
         website: "",
         organizations: [],
+        funding: []
       },
       // TODO: should this come from the backend?
       access: {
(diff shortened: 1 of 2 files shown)


invenio_communities/webpack.py
```diff
@@ -54,6 +54,8 @@
                 '@ckeditor/ckeditor5-react': '^2.1.0',
                 '@semantic-ui-react/css-patch': '^1.0.0',
                 'react-invenio-forms': '^0.10.0',
+                # TODO update after releasing Funding Field
+                'react-invenio-deposit': '^0.19.4',
```
alejandromumo, line 58: This should be setup to the correct version after https://github.com/inveniosoftware/react-invenio-deposit/pull/465/files gets released

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -256,18 +258,124 @@ class CommunityProfileForm extends Component {
...
     });
 
+    /**
+     * Deserializes a funding record (e.g. funder or award)
+     *
+     * @param {object} fund
+     *
+     * @returns {object} an object containing the deserialized record
+     */
+     const deserializeFunding = ((fund) => {
```
alejandromumo, line 268: This is exactly the same logic to deserialise a funding record that was used in  https://github.com/inveniosoftware/react-invenio-deposit/blob/aefebaa2156c752ee10829d69c85b3ab22a15eeb/src/lib/fields/VocabularyField.js#L42-L44

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -256,18 +258,124 @@ class CommunityProfileForm extends Component {
...
   serializeValues = (values) => {
+    /**
+     * Serializes a funding record (e.g. funder or award)
+     *
+     * @param {object} fund
+     *
+     * @returns {object} an object containing the serialized record
+     */
+     const serializeFunding = ((fund) => {
+      const _serialize = (value) => {
```
alejandromumo, line 343: This is exactly the same logic to serialise a funding record that was used in

https://github.com/inveniosoftware/react-invenio-deposit/blob/aefebaa2156c752ee10829d69c85b3ab22a15eeb/src/lib/fields/VocabularyField.js#L296

zzacharo: Can you just import the field and use the serialize method? Maybe it's not exported but we could export the field eventually....

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -426,27 +541,106 @@ class CommunityProfileForm extends Component {
...
                     allowAdditions
                     search={(filteredOptions, searchQuery) => filteredOptions}
                   />
+                  <FundingField
+                    fieldPath="metadata.funding"
+                    searchConfig={{
+                      searchApi: {
+                        axios: {
+                          headers: {
+                            //  FIXME use for internationalisation
```
zzacharo, line 559: Shall we fix that? Or create an issue and delete the comment?

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -432,6 +550,71 @@ class CommunityProfileForm extends Component {
...
+                    }
+                    computeFundingContents={(funding) => {
+                      let headerContent, descriptionContent = '';
+                      let awardOrFunder = 'award';
+                      if (funding.award) {
+                        headerContent = funding.award.title;
+                      }
+
+                      if (funding.funder) {
+                        const funderName = funding.funder?.name ?? funding.funder?.title?.en ?? funding.funder?.id ?? '';
```
zzacharo, line 606: Shouldn't the i18n title come first? Is the funder a person or organization type? If so, isn't it better to check explicitly what type of funder you have and be stricter with where to check for the name?

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
zzacharo: Can you just import the field and use the deserialize method? Maybe it's not exported but we could export the field eventually....

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -432,6 +550,71 @@ class CommunityProfileForm extends Component {
...
+                    fieldPath="metadata.funding"
+                    searchConfig={{
+                      searchApi: {
+                        axios: {
+                          headers: {
+                            //  FIXME use for internationalisation
+                            //  Accept: "application/vnd.inveniordm.v1+json"
+                            Accept: "application/json",
+                          },
+                          url: "https://127.0.0.1:5000/api/awards",
```
zzacharo, line 563: 👀

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
alejandromumo: I think the backend for funders and grants is not ready yet to internationalise (e.g. `funder.title` should be declared as a `L10NString`).

PR #583, opened 2022-05-04
global: pid to slug model field

lnielsen:
- Changes from using a persistent identifier for the user decided identifier to using a model slug field.
- Removes left-overs from old owners field.


ntarocco:
Needs rebase: when rebasing, you can probably skip the incoming changes (add `uuid` to the mapping and the `read_many` service method, they don't make sense anymore here...)

Issue #459, opened 2022-03-11
Members UI: assumptions to verify after backend implementation


jrcastro2:
This a list of assumptions made during the development of the UI that need to be tested upon completion:
- [x] In the memeber list view as a manager/owner/member, the actions that can be done are determined by the backend, this includes:
  - [x] Verify "Visibility" rules
  - [x] Verify "Role" rules
  - [x] Verify "Remove" and "Leave" actions
- [x] Verify facets

PR #601, opened 2022-05-05
members: fix permission check

jrcastro2:
* Fix filters label alignment
* Fix display of boolean values in filters
* Pass filters value from the backend
* closes https://github.com/inveniosoftware/invenio-communities/issues/459

image

## As a manager view
image

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -169,3 +169,46 @@ COMMUNITIES_INVITATIONS_SORT_OPTIONS = {
 
 COMMUNITIES_INVITATIONS_EXPIRES_IN = timedelta(days=30)
 """"Default amount of time before an invitation expires."""
+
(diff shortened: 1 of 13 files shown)


invenio_communities/config.py
```diff
@@ -169,3 +169,46 @@
...
+    },
+    "visibility": {
+        "label": "Visibility",
+        "filterValues": [
+            {"label": "Hidden", "value": "false"},
+            {"label": "Public", "value": "true"},
+        ]
+    }
+}
+""""Community members filters to display in the UI."""
```
jrcastro2, line 214: We agreed (some time ago) that this should come from the backend but not sure this structure is the best, any better suggestion are welcomed!

jrcastro2:
Closed in favour of https://github.com/inveniosoftware/invenio-communities/pull/605

PR #602, opened 2022-05-05
Members: update data modified by the bulk actions

Pineirin:
closes: inveniosofware/invenio-communities#556


invenio_communities/members/services/service.py
```diff
@@ -415,6 +422,9 @@ def update(self, identity, community_id, data, uow=None):
                     _("A community must have at least one owner.")
                 )
 
+        if refresh == True:
```
ntarocco, line 425:
```suggestion
        if refresh:
```

Pineirin: Ups, I was with a js mentality.

PR #634, opened 2022-05-10
Members: remove checkbox from current user [DEPENDS ON https://github.com/inveniosoftware/invenio-theme/pull/283]

Pineirin:
* closes: #629

Screenshot from 2022-05-10 11-55-33
Screenshot from 2022-05-10 11-55-55
Screenshot from 2022-05-10 11-56-13


ntarocco (approved):

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js
```diff
@@ -52,10 +52,16 @@ export class ManagerMembersResultItem extends Component {
     return (
       <Table.Row>
         <Table.Cell className="selected-member">
-          <Grid textAlign="left" verticalAlign="middle">
+          <Grid
+            textAlign="left"
+            verticalAlign="middle"
+            className={result.is_current_user && "current-user"}
```
kpsherva, line 58: this class should be set on `<Item` that contains the checkbox, not the whole grid. There is no reason to alter the whole grid
the naming of the class should be more descriptive (following semantic ui principles), on what is actually doing. the class is not describing the behaviour of current user (because we don't display it) but adding margin/padding to fill the removed checkbox space. I would go with 'align-no-checkbox` ?

kpsherva (approved):

Issue #627, opened 2022-05-10
add `COMMUNITIES_ALLOW_GROUP_INVITES` to backend


ntarocco:
1. rename `COMMUNITIES_ALLOW_GROUP_INVITES` to something like `COMMUNITIES_ENABLE_GROUPS`, ask for default value
2. add the config var in the backend **as a permission** check

PR #648, opened 2022-05-11
Members invitations: rename COMMUNITIES_ALLOW_GROUP_INVITES flag

Pineirin:
Ensure the above mentioned flag is enable when adding members in the backend.
closes: #627

Testing of the workflow of inviting a user.
Screenshot from 2022-05-13 12-34-21
Screenshot from 2022-05-13 12-34-56
Screenshot from 2022-05-13 12-35-08
Screenshot from 2022-05-13 12-35-26
Screenshot from 2022-05-13 12-35-41
Screenshot from 2022-05-13 12-35-50
Screenshot from 2022-05-13 12-36-02
Screenshot from 2022-05-13 12-36-10
Screenshot from 2022-05-13 12-36-26

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -28,7 +28,7 @@ COMMUNITIES_ROUTES = {
 
 """Communities ui endpoints."""
 
-COMMUNITIES_ALLOW_GROUP_INVITES = True
+COMMUNITIES_GROUPS_ENABLED = True
 """Config to allow invitation of groups."""
 
 COMMUNITIES_FACETS = {
--- a/invenio_communities/generators.py
+++ b/invenio_communities/generators.py
@@ -17,6 +17,7 @@ from functools import partial, reduce
 from itertools import chain
 
 from elasticsearch_dsl.query import Q
+from flask import current_app
 from flask_principal import UserNeed
 from invenio_access.permissions import any_user, system_process
 from invenio_records_permissions.generators import Generator
@@ -276,5 +277,27 @@ class AllowedMemberTypes(Generator):
         if member_types:
             for m in member_types:
                 if m not in self.allowed_member_types:
+                    return [any_user]
+        return []
+
+class GroupsEnabled(Generator):
+    """Generator to restrict if the groups are not enabled.
+
+    If the groups are not enabled, exclude any user for adding members of the
+    param member type.
+
+    A system process is allowed to do anything.
+    """
+
+    def __init__(self, *need_groups_enabled_types):
+        """Types that need the groups enabled."""
+        self.need_groups_enabled_types = need_groups_enabled_types
+
+    def excludes(self, member_types=None, **kwargs):
+        """Preventing needs."""
+        if member_types:
+            for m in member_types:
+                if m in self.need_groups_enabled_types and \
+                    not current_app.config["COMMUNITIES_GROUPS_ENABLED"]:
                     return [any_user]
         return []
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationsSearchLayout.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationsSearchLayout.js
@@ -21,7 +21,7 @@ export class InvitationsSearchLayout extends Component {
       roles,
       rolesCanInvite,
       community,
-      communityAllowGroupInvites,
+      communityGroupsEnabled,
     } = this.props;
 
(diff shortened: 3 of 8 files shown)


invenio_communities/generators.py
```diff
@@ -275,6 +276,9 @@ def excludes(self, member_types=None, **kwargs):
         """Preventing needs."""
         if member_types:
             for m in member_types:
+                if m=='group' and \
```
kpsherva, line 279: is there any way to avoid hardcoding `group` in this condition statement ? ping @ntarocco

tests/members/test_members_services.py
```diff
@@ -68,6 +69,33 @@ def test_add_denied(
         data
     )
 
+@pytest.mark.parametrize("actor,role", [
```
ppanero, line 72: This is good, and very comprehensive 💪. Can we add another test, with just one use case:
- Having the flag enabled (with the fixture mentioned below)
- Adding a user (not group), to test that the permissions work properly (i.e. that they dont deny by default).

ppanero, line 72: Using `parametrize` is very nice and clean. However, it has the downside of calling the function once per use case (in this case 7 times). When the fixtures used by the test function are of scope "module" it is okay to use `parametrize` since fixtures will only be created/destroyed once. However, this test has `members` which creates serveral users (4). Those users are being created/removed each time (i.e. 28 calls + db session destroy/rollback).

I merely want to avoid ending up in a situation like rdm-records, where tests take 10+minutes to run.

Can this test be refactored with a for loop inside the test function?

ppanero (commented):
Nice work! Still need to test UI wise, but you can go ahead implementing the changes 🚀

invenio_communities/generators.py
ppanero: +1 to this

ppanero, line 279:
```suggestion
                if m == 'group' and \
```

ntarocco: we discussed this with @Pineirin, when called, it is passing a string `group`... so the `AllowedMemberTypes` is already quite coupled with `user`/`email`/`group` strings.
Another way is to create a totally new permission class (as he did before), e.g. `VerifyGroupEnabled` or whatever, but if you override `can_members_add` in your instance and you don't add that class, then it will not work.
One or the other...

Pineirin: I don't think it would make sense to use a configurable value in the generator and not in the permission itself. I would either use types defined in the configuration for all AllowedMemberTypes. In case two, I'm ok with it but I don't know if the con Nicola mentioned is too big.

I don't have enough confidence to state that any solution better (as Nicola said, I've been bouncing between both).

Pineirin: It was decided IRL, that we would use the second solution.

tests/members/test_members_services.py
Pineirin: The answer to this comment was posted internally on mattermost due to it's length.

invenio_communities/config.py
```diff
@@ -28,7 +28,7 @@
 
 """Communities ui endpoints."""
 
-COMMUNITIES_ALLOW_GROUP_INVITES = True
+COMMUNITIES_ENABLED = True
```
kpsherva, line 31: I am confused on this one. I thought we are removing `COMMUNITIES_ENABLED` ? ping @ntarocco

invenio_communities/generators.py
```diff
@@ -278,3 +307,25 @@ def excludes(self, member_types=None, **kwargs):
...
+        self.need_groups_enabled_types = need_groups_enabled_types
+
+    def excludes(self, member_types=None, **kwargs):
+        """Preventing needs."""
+        if member_types:
+            for m in member_types:
+                if m in self.need_groups_enabled_types and \
+                    not current_app.config["COMMUNITIES_GROUPS_ENABLED"]:
+                    return [any_user]
+        return []
```
kpsherva, line 331: missing new line

tests/members/test_members_services.py
ppanero: Can you add the answer here. If someone else has the same doubt they cannot see the mattermost chat ;)

invenio_communities/config.py
ppanero: As far as I understood we will have communities enabled (not allowed to disabled), however groups can be enabled/disabled.

```suggestion
COMMUNITIES_GROUPS_ENABLED = True/False
```

PR #387, opened 2022-02-02
global: enable communities by default

lnielsen:

PR #654, opened 2022-05-12
global: enable communities by default

ntarocco:


ntarocco:
supersedes https://github.com/inveniosoftware/invenio-communities/pull/387

PR #663, opened 2022-05-16
community header: layout improvements

jennur:
**Needs** https://github.com/inveniosoftware/invenio-app-rdm/pull/1574

Reduced the amount of space needed for the header content.
- Moved description to accordion
- Changed details listing to be inline on computer (needs PR mentioned above - use of `.flex` does not solve the issue, as this class does not break the content on smaller screens.)
- Made logo slightly smaller on tablet

## Screenshots
Screenshot 2022-05-16 at 14 43 20
Screenshot 2022-05-16 at 14 44 24
Screenshot 2022-05-16 at 14 43 32
Screenshot 2022-05-16 at 14 11 59

Screenshot 2022-05-16 at 14 45 15
Screenshot 2022-05-16 at 14 45 25
Screenshot 2022-05-16 at 14 12 50

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
@@ -12,47 +12,68 @@
 <div class="ui container fluid page-subheader-outer with-submenu rel-pt-2">
   <div class="ui container relaxed grid page-subheader">
     <div class="row pb-0">
-      <div class="sixteen wide mobile five wide tablet two wide computer column">
-        <div class="ui rounded fluid image community-image rel-mb-2">
+      <div class="sixteen wide mobile sixteen wide tablet thirteen wide computer column">
+
+        <div class="ui rounded image community-image mt-5 rel-mr-2 left-floated">
           <img src="{{ community.links.logo }}" alt="Community logo" onerror="this.onerror=null;this.src='/static/images/square-placeholder.png';"/>
         </div>
-      </div>
 
-      <div class="sixteen wide mobile eleven wide tablet eleven wide computer column rel-mb-2">
-        <h2 class="ui header">{{ community.metadata.title }}</h2>
-        <div class="rel-mb-2">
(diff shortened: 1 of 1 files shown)


kpsherva:
nice idea with "About this community" button! would it make sense to be always "open" on desktop ? and folded when there is not enough space?

invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -12,45 +12,65 @@
...
+          {% if community.metadata.organizations %}
+          <div class="inline-computer mt-5 rel-mr-1">
+            <i class="building outline icon" aria-hidden="true"></i>
+            <span class="label label-keyword">{{ community.metadata.organizations| map(attribute='name') | join(', ') }}</span>
+          </div>
+          {% endif %}
         </div>
+
+        {% if community.metadata.description %}
+          <div class="ui accordion mt-5">
```
kpsherva, line 52: I would suggest to use something more like "show more/show less" button instead of the accordion, because i am not sure if we want to hide the description entirely

jennur: So to keep the same as before you mean?

lnielsen:
Hey

Over all I'd like to significantly reduce the header size. Goal issomething like GitHub as an example:

Screenshot 2022-05-17 at 14 00 42

Thus,

- I would reduce logo size even further
- Agree that description should be default be hidden
- New upload button is ok to vertically centre when the entire header size is reduced:

168595141-7387da30-1204-4412-aa4e-8c9d90555889

jennur:
**NOTE: Updates needs PR https://github.com/inveniosoftware/invenio-app-rdm/pull/1606 for styling!**

## Updated screenshots
Screenshot 2022-05-17 at 16 01 17
Screenshot 2022-05-17 at 16 01 26
Screenshot 2022-05-17 at 16 01 35
Screenshot 2022-05-17 at 16 01 42

PR #669, opened 2022-05-17
members: remove link from public member

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1591

Screenshot from 2022-05-17 13-48-54


Pineirin (approved):
:+1:

jrcastro2:
I would try to highlight the name (making it bold, could be an idea), otherwise in the screenshot above it looks like the full name is "John Doe Cern". The point is to differentiate the name from the affiliation/description, WDYT?

NRodriguezcuellar:
> I would try to highlight the name (making it bold, could be an idea), otherwise in the screenshot above it looks like the full name is "John Doe Cern". The point is to differentiate the name from the affiliation/description, WDYT?

makes sense!

kpsherva (approved):

PR #672, opened 2022-05-17
inivitation: fix "jumping" of modal

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1600
* needs https://github.com/inveniosoftware/invenio-app-rdm/pull/1605

https://user-images.githubusercontent.com/15194802/169237992-f4a0504e-e2d1-4139-aa0b-cf2750b802a4.mp4

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/components/bulk_actions/SelectedMembers.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/components/bulk_actions/SelectedMembers.js
@@ -25,14 +25,11 @@ export class SelectedMembers extends Component {
 
     return !_isEmpty(selectedMembers) ? (
       <>
-        <Header as="h2" size="small">
-          {i18next.t(displayingGroups ? "Selected groups" : "Selected members")}
-        </Header>
         <Segment className="selected-members-header mb-20 mr-20">
           {Object.entries(selectedMembers).map(([memberId, member]) => (
             <Button
               key={memberId}
-              className="p-0 mr-10"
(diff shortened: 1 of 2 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/members/components/bulk_actions/SelectedMembers.js
```diff
@@ -49,17 +46,15 @@ export class SelectedMembers extends Component {
...
+          <Icon name="users" />
 
-            {i18next.t(
-              displayingGroups ? "No groups selected." : "No people selected."
-            )}
-          </Header>
-        </Segment>
-      </div>
+          {i18next.t(
+            displayingGroups ? "No groups selected." : "No people selected."
```
kpsherva, line 54: translations missing

kpsherva:
in the video there is still change of the height of the modal visible when you are changing tabs, was this the "jump"?
if you select more members (> 5, 10) will the selected members area have a scroll?

jrcastro2:
> in the video there is still change of the height of the modal visible when you are changing tabs, was this the "jump"? if you select more members (> 5, 10) will the selected members area have a scroll?

The "jump" was referring to the change of height of the selected members when selected a member for the fist time. Now the placeholder for "No people selected" has the same height as the initial selected members area. After discussing about the jump when changing tabs during the stand-up we agreed to also add the "Message" text area to the group tab and disable it, to avoid the jumping. 
When we select many members the area will expand, there is no scroll, this behaviour was confirmed in the stand-up.

zzacharo (approved):

PR #681, opened 2022-05-19
Invitations: fix tablet view [DEPENDENCIES]

Pineirin:
fix mobile view.
add status icon.
add extra space for user full name.
closes: inveniosoftware/invenio-app-rdm#1595
depends on: https://github.com/inveniosoftware/invenio-app-rdm/pull/1640
depends on: https://github.com/inveniosoftware/invenio-requests/pull/238

Screenshot from 2022-05-23 14-59-42
Screenshot from 2022-05-23 14-59-50
Screenshot from 2022-05-23 14-59-58

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationResultItem.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationResultItem.js
@@ -14,8 +14,8 @@ import { Image } from "react-invenio-forms";
 import { Container, Grid, Item, Table } from "semantic-ui-react";
 import { InvitationsContext } from "../../api/invitations/InvitationsContextProvider";
 import { RoleDropdown } from "../components/dropdowns";
-import _capitalize from "lodash/capitalize";
 import Overridable from "react-overridable";
+import RequestStatus from "@js/invenio_requests/request/RequestStatus";
 
 const formattedTime = (expires_at) =>
   DateTime.fromISO(expires_at).setLocale(i18next.language).toRelative();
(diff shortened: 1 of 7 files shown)


jennur (commented):
- The status and the icon should be aligned on mobile.
- Maybe also align the text with the avatar above.
- The issue also mentions the unused white space on desktop. Maybe we can give even more space for the name on desktop, as the names can potentially be very long? And make sure that there is not an empty column at the end?
 What do you think?

For mobile and tablet:
Screenshot 2022-05-20 at 09 56 34

Pineirin:
Regarding the last point. I've defined the width of the element dedicated to displaying the full name to be the same than the one in the members view (6/16). Is true that I've left some space (2/16) free at the right, but that was reserved for when the links are available for the invitations requests.

Screenshot from 2022-05-23 14-36-50

 The first two points where properly addressed as can be seen in the screenshots.

ntarocco (approved):
LGTM! Just added a comment about potential code duplication

invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/RequestStatus.js
```diff
@@ -0,0 +1,51 @@
...
+// Invenio RDM Records is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React from "react";
+import RequestStatusIcon from "@js/invenio_requests/request/RequestStatusIcon";
+import { i18next } from "@translations/invenio_communities/i18next";
+
+export const SubmitStatus = () => (
+  <div>
+    <RequestStatusIcon status="submitted" />
```
kpsherva, line 13: status should not be hardcoded in the prop, but it should be passed from the parent component

invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationResultItem.js
```diff
@@ -80,13 +80,27 @@ export class InvitationResultItem extends Component {
...
           </Grid>
         </Table.Cell>
-        <Table.Cell>
+        <Table.Cell
+          className="mobile-table-cell"
+          data-label={i18next.t("Status")}
+        >
           <Overridable id={`RequestStatus.${request.status}`}>
-            <span>{request.status}</span>
+            <div>
```
kpsherva, line 88: this should be a separate component `RequestStatus` with it's own Overridable id, not "hidden" in another component, like it is done in line 87.
I would suggest renaming existing `RequestStatusIcon` to `RequestStatus` and extending it to contain the status name as well, instead of adding 5 new overridden components.

kpsherva (approved):

Issue #690, opened 2022-05-30
communities: members search checkbox misaligned


kpsherva:
<img width="1376" alt="Screenshot 2022-05-30 at 16 16 35" src="https://user-images.githubusercontent.com/38131488/171011059-15c98b4a-8d2c-4687-ac37-cef01a5f894d.png">

PR #691, opened 2022-05-30
members: fix checkbox and avatar alignment

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-communities/issues/690

image


invenio_communities/assets/semantic-ui/js/invenio_communities/members/components/bulk_actions/SearchResultsRowCheckbox.js
```diff
@@ -46,13 +46,15 @@ export class SearchResultsRowCheckbox extends Component {
...
     const { bulkActionContext, allSelected } = this.context;
+    const { className } = this.props;
     return (
       <>
         <Checkbox
           checked={
             this.isChecked(bulkActionContext, allSelected) || allSelected
           }
           onChange={this.handleOnChange}
+          className={className && className}
```
kpsherva, line 57:
```suggestion
          className={className}
```

invenio_communities/assets/semantic-ui/js/invenio_communities/members/components/bulk_actions/SearchResultsRowCheckbox.js
```diff
@@ -61,4 +63,9 @@ export class SearchResultsRowCheckbox extends Component {
 
 SearchResultsRowCheckbox.propTypes = {
   rowId: PropTypes.string.isRequired,
+  className: PropTypes.string,
+};
+
+SearchResultsRowCheckbox.defaultProps = {
+  className: undefined,
```
kpsherva, line 70:
```suggestion
  className: '',
```

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js
```diff
@@ -75,9 +75,9 @@ export class ManagerMembersResultItem extends Component {
                 key={result.id}
               >
                 {!result.is_current_user && (
-                  <SearchResultsRowCheckbox rowId={result.id} data={result} />
+                  <SearchResultsRowCheckbox className="mt-10" rowId={result.id} data={result} />
```
kpsherva, line 78: could you provide a screenshot of how it works with more items on the list ?

jrcastro2: Sure, it would look like this
image

kpsherva: I am asking in details, since the "flex" class should take care of the alignment, instead of adding margins to the image. Especially because the margin is not in relative units. I think the flex class is somehow broken or not applied properly, due to some changes. I would suggest investigating on that instead of applying the margin because margin is only a workaround in this case

jrcastro2: It looks good in all devices I tested, see:

## Ipad air
127 0 0 1_5000_communities_123_members_q= l=list p=1 s=25 sort=name(iPad Air)

## Iphone SE
127 0 0 1_5000_communities_123_members_q= l=list p=1 s=25 sort=name(iPhone SE)


But I can give it a try to get rid of the margins.

jrcastro2: After testing I am not sure we can get rif of the margins I applied to the logo the following rule to center it `align-self: center;` and remove the margins but the checkbox is still not well centered, see image:
image

The checkbox has some margins inherited from [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/views/item.overrides#L123-L127) but the margin of 1em set there is too much. If I am not mistaken the grid size is static, personally I think setting the margins is OK in this case as it allows to align it correctly, see image:
image

jrcastro2:
Closed in favour of https://github.com/inveniosoftware/invenio-app-rdm/pull/1684 as the change should be done in the theming as discussed above

PR #704, opened 2022-06-07
[NEEDS #invenio-vocabularies/pull/199 #invenio-rdm-records/pull/979]recources: add ui serializer

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1518
* needs https://github.com/inveniosoftware/invenio-rdm-records/pull/979

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
@@ -290,8 +290,8 @@ class CommunityProfileForm extends Component {
       const _deserialize = (value) => {
         const deserializedValue = _cloneDeep(value);
 
-        if (value?.title?.en) {
-          deserializedValue.title = value.title.en;
+        if (value?.title_l10n) {
+          deserializedValue.title = value.title_l10n;
         }
 
         if (value.identifiers) {
@@ -325,7 +325,7 @@ class CommunityProfileForm extends Component {
       return deserializedValue;
     };
 
-    const funding = initialValues.metadata.funding.map((fund) => {
+    const funding = initialValues.ui.funding?.map((fund) => {
       return {
         ...(fund.award && { award: deserializeFunding(fund.award) }),
         funder: deserializeFunding(fund.funder),
@@ -390,7 +390,6 @@ class CommunityProfileForm extends Component {
           ? fund.map(_serialize)
           : _serialize(fund);
       }
-
       return serializedValue;
     };
 
@@ -406,9 +405,8 @@ class CommunityProfileForm extends Component {
         };
       }
     );
-
     // Serialize each funding record, award being optional.
-    const funding = submittedCommunity.metadata?.funding.map((fund) => {
+    const funding = submittedCommunity.metadata?.funding?.map((fund) => {
       return {
         ...(fund.award && { award: serializeFunding(fund.award) }),
         funder: serializeFunding(fund.funder),
@@ -454,8 +452,8 @@ class CommunityProfileForm extends Component {
       }
     }
   };
-
   render() {
+    const { types } = this.props;
     return (
       <Formik
         initialValues={this.getInitialValues(this.props.community)}
@@ -505,10 +503,10 @@ class CommunityProfileForm extends Component {
                         label={i18next.t("Type")}
                       />
                     }
-                    options={this.props.types.map((ct) => {
+                    options={types.map((ct) => {
                       return {
                         value: ct.id,
-                        text: ct?.title?.en ?? ct.id,
+                        text: ct?.title_l10n ?? ct.id,
                       };
                     })}
                   />
@@ -579,9 +577,7 @@ class CommunityProfileForm extends Component {
                       searchApi: {
                         axios: {
                           headers: {
-                            //  FIXME use for internationalisation
-                            //  Accept: "application/vnd.inveniordm.v1+json"
-                            Accept: "application/json",
+                            Accept: "application/vnd.inveniordm.v1+json",
                           },
                           url: "/api/awards",
                           withCredentials: false,
@@ -599,7 +595,7 @@ class CommunityProfileForm extends Component {
                     labelIcon="money bill alternate outline"
                     deserializeAward={(award) => {
                       return {
-                        title: award.title.en ?? award.title,
+                        title: award.title_l10n,
                         pid: award.pid,
                         number: award.number,
                         funder: award.funder ?? "",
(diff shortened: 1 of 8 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -423,6 +423,9 @@ class CommunityProfileForm extends Component {
     // Clean values
     submittedCommunity = removeEmptyValues(submittedCommunity);
 
+    // Remove ui object
+    delete submittedCommunity.ui;
```
kpsherva, line 427: isn't it better to make the `.ui` key dump_only and let the backend handle it? it is going to be quite problematic to always remember to remove it in the frontend

invenio_communities/views/decorators.py
```diff
@@ -11,18 +11,33 @@
...
+        @wraps(f)
+        def view(**kwargs):
+            pid_value = kwargs["pid_value"]
+            community = current_communities.service.read(
+                id_=pid_value, identity=g.identity
+            )
+            kwargs["community"] = community
+            if serialize:
+                community_ui = CustomJSONSerializer(
+                    object_key='ui',
```
kpsherva, line 34: I find it a bit complex to specify the "ui" key on each usage, I had an impression we were trying to "standarize" it across the code base. How about having a new class inherited from `MarshmallowSerializer` which assigns these 4 arguments as class attributes? 
Is it possible to have it registered with the community record in the config? ping @ntarocco what do you think?

kpsherva: after re-reading the PR, I see the response serializer is already registered in the resource config, but using it here in the decorator actually prevents the config from being used, I think it would be better to use the response serializer rather than implementing it here in the decorator

invenio_communities/views/decorators.py
```diff
@@ -11,18 +11,33 @@
...
 from flask import g
+from flask_resources import CustomJSONSerializer, JSONSerializer, UIListSchema
 
+from invenio_communities.communities.resources.UISchema import \
+    UICommunitySchema
 from invenio_communities.proxies import current_communities
 
 
-def pass_community(f):
+def pass_community(serialize=False):
```
ntarocco, line 21: I would avoid the default value `False`: this is because when using the decorator, if you don't pass the param (because your want it `False`), Flask will not be happy and will throw an error. It took me quite some time last time to debug it.
TL;DR: when a decorator as params, it must be always be called with the params.

invenio_communities/views/decorators.py
ntarocco: The one configured in the `config` is for REST APIs, I am not sure that you can easily reuse it here.
In any case, maybe we can indeed create a `CommunitySerializer` or similar, instantiating (or inheriting if not possible to compose) `MarshmallowSerializer` with the params here.

invenio_communities/views/decorators.py
jrcastro2: Ok, thanks a lot for sharing!

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
jrcastro2: I am not 100% sure about this, I understand the point that it can be a bit more problematic but I think it should be the client responsible for sending some valid data and not the backend adapting to the client needs, note that the schema used to fetch the data is different than the one used to post the data, this is due to localization as we need to translate the vocabularies. That said we could easily add to the community Marshamallow schema `unknown = EXCLUDE` but this implies that when using the API we will not get an error if we add non-existing fields, I am fine with this, if this works for you I could add this change but want to double check with you first.

Please note that for the records we do also remove the `ui` object before sending the data  (see code [here](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositRecordSerializer.js#L362-L370)). If we want to change this behaviour I think that we should do it for both. 

I am also available to discuss IRL if it's easier.

ntarocco: @slint @alejandromumo any idea why the `unknown = EXCLUDE` was not added? Simply forgotten?

slint: Probably yes, it was forgotten... This was one of the very first forms we did in React, so the common services/resources/schemas practices we have now were not that common.

invenio_communities/communities/resources/serializer.py
```diff
@@ -0,0 +1,24 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio Communities Resource Serializers."""
+
+from flask_resources import MarshmallowSerializer, UIListSchema, JSONSerializer
+
+from invenio_communities.communities.resources.UISchema import UICommunitySchema
+
+
+class UICommunityJSONSerializer(MarshmallowSerializer):
```
zzacharo, line 16: nit: I would rename it to `CommunityUIJSONSerializer`. Not blocking :)

invenio_communities/views/decorators.py
```diff
@@ -12,17 +12,31 @@
...
-
-    return view
+    def decorator(f):
+        @wraps(f)
+        def view(**kwargs):
+            pid_value = kwargs["pid_value"]
+            community = current_communities.service.read(
+                id_=pid_value, identity=g.identity
+            )
+            kwargs["community"] = community
```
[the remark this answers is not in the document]
jrcastro2: Yes, we need to pass the `community` too because we also need to check for the permissions, see [here an example](https://github.com/inveniosoftware/invenio-communities/blob/5f46f455ccac721489b33e88b2877d07ae85feaa/invenio_communities/views/communities.py#L98).

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -406,14 +405,15 @@ class CommunityProfileForm extends Component {
...
-
     // Serialize each funding record, award being optional.
-    const funding = submittedCommunity.metadata?.funding.map((fund) => {
-      return {
-        ...(fund.award && { award: serializeFunding(fund.award) }),
-        funder: serializeFunding(fund.funder),
-      };
-    });
+    const funding =
+      submittedCommunity.metadata?.funding &&
```
zzacharo, line 410: this is redundant as we use the `?.`...is there an issue?

jrcastro2: thanks!

zzacharo (approved):

PR #706, opened 2022-06-08
communities logo: set height to avoid breaking the layout

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1704
needs https://github.com/inveniosoftware/invenio-app-rdm/pull/1710
Screenshot from 2022-06-13 13-59-23

Screenshot from 2022-06-13 13-59-32

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
@@ -151,6 +151,7 @@ const LogoUploader = ({ community, defaultLogo, hasLogo, onError }) => {
               fluid
               wrapped
               rounded
+              className="community-logo settings"
             />
 
             <Divider hidden />


invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -151,6 +151,7 @@ const LogoUploader = ({ community, defaultLogo, hasLogo, onError }) => {
               fluid
               wrapped
               rounded
+              className="community-logo tall"
```
kpsherva, line 154: I am not sure what is the reason to add a separate `tall` class to limit the height ? why not to do it simply with `community-logo` ?

NRodriguezcuellar: because there is more space in the settings tab to show the logo. the height will be too big for the list items

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -151,6 +151,7 @@ const LogoUploader = ({ community, defaultLogo, hasLogo, onError }) => {
               fluid
               wrapped
               rounded
+              className="tall community-logo"
```
kpsherva, line 154: I find this class name misleading. How do you know the image will be tall ?

NRodriguezcuellar: `tall` is a relative term of course, How i would interpret it is that this logo will be taller than all other community logos in the application, but you are right that, as this is only used on one page (and probably will stay like that) i could make it specific to the settings page.

jennur (approved):

PR #707, opened 2022-06-08
communities: prevent errors in type field

NRodriguezcuellar:
type attribute now checks if the value passed is a dict to prevent errors from happening before the value is validated. Now returns a normal validation error if the wrong type of value is passed.

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1699

Screenshot from 2022-06-08 11-38-07
Screenshot from 2022-06-08 11-36-57


invenio_communities/communities/schema.py
```diff
@@ -75,9 +75,10 @@ def clean(self, data, **kwargs):
         Why: We want to allow the output of a Schema dump, to be a valid input
              to a Schema load without causing strange issues.
         """
-        for name, field in self.fields.items():
-            if field.dump_only:
-                data.pop(name, None)
+        if type(data) is dict:
```
slint, line 78:
```suggestion
        if isinstance(data, dict):
```

Minor: See [here](https://switowski.com/blog/type-vs-isinstance#:~:text=Difference%20between%20isinstance%20and%20type&text=type%20only%20returns%20the%20type,specified%20as%20a%20second%20parameter.) why :)

slint (commented):
Minor comment for better type checking. If possible, can you also check if there's a test for this that validates the `type` field?

Otherwise LGTM :)

NRodriguezcuellar:
> Minor comment for better type checking. If possible, can you also check if there's a test for this that validates the `type` field?
> 
> Otherwise LGTM 
There is a test to check if the value of the type `id` is valid here https://github.com/inveniosoftware/invenio-communities/blob/master/tests/communities/test_relations_types.py.  Not necessarily anything testing what i am doing here.

PR #724, opened 2022-06-20
global: Fixes strings marked for translation.

chriz-uniba:
- Removes superfluous white space.
- Adjust field to look like other fields.

Refs: inveniosoftware/invenio-app-rdm#1707

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/RenameCommunitySlugButton.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/RenameCommunitySlugButton.js
@@ -90,7 +90,8 @@ export class RenameCommunitySlugButton extends Component {
           <Modal.Content>
             <Form onSubmit={this.handleChange}>
               <Form.Input
-                label={i18next.t("Enter the new unique identifier of the community.")}
+                label={i18next.t("New unique identifier of the community")}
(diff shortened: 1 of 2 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/RenameCommunitySlugButton.js
```diff
@@ -90,7 +90,8 @@ export class RenameCommunitySlugButton extends Component {
           <Modal.Content>
             <Form onSubmit={this.handleChange}>
               <Form.Input
-                label={i18next.t("Enter the new unique identifier of the community.")}
+                label={i18next.t("New unique identifier of the community")}
+                placeholder={i18next.t("New unique identifier of the community")}
```
kpsherva, line 94: could you show us how does it look like (screenshot)? I am not sure if the field label should be the same as a placeholder

chriz-uniba: image

Popup shown when clicking here on "Change identifier"
image

chriz-uniba: I thought it might be unique to other places like this:
image
and this
image
however, both of them aren't pop-ups.

PR #726, opened 2022-06-21
[NEEDS #invenio-search-ui/pull/141] Multiple searchbar

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1701


alejandromumo (approved):

invenio_communities/templates/semantic-ui/invenio_communities/header.html
```diff
@@ -0,0 +1,34 @@
...
+    {% set defaultOption = 
+        {
+        "key": "records",
+        "text": _("All") + " " + config.THEME_SITENAME,
+        "value": config.APP_RDM_ROUTES.record_search,
+        }
+    %}
+    {% set options = [defaultOption] %}
+    {% if community %}
+        {% set community_url = config.RDM_COMMUNITIES_ROUTES["community-detail"].replace("<pid_value>", community.slug) %}
```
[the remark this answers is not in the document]
jrcastro2: `Community` is injected in the jinja template, the value is coming from the DB.

ntarocco (approved):
LGTM!

invenio_communities/templates/semantic-ui/invenio_communities/header.html
```diff
@@ -0,0 +1,34 @@
...
+        {% set communityOption = 
+        {
+            "key": "communities",
+            "text": _("In this community"),
+            "value": community_url,
+        }
+        %}
+        {% set options = [communityOption, defaultOption] %}
+    {% endif %}
+    {{ nav_searchbar(options) }}
```
kpsherva, line 33: this macro is coming from invenio-app-rdm and unfortunately it cannot be used in here. If for some reason we install invenio-communities to any other app (ils?) this will fail since the macro will not be defined

jrcastro2: Right! 
Would it make sense to move the macro to `invenio-theme`? or any better proposals?

kpsherva: in this case we would have to move the entire searchbar to invenio-theme, otherwise the macro would not have a reason to exist there. I simplified the implementation for now, it can be rechallenged though

Pineirin (approved):

Issue #732, opened 2022-06-27
communities: add `restricted` label to the community header if the community is restricted


kpsherva:
- [x] add `restricted` label in community header
- [x] add more information in the community selection modal (deposit)
<img width="958" alt="Screenshot 2022-06-20 at 11 18 29" src="https://user-images.githubusercontent.com/38131488/175896100-37941787-55fa-4f42-a4c3-fe54c4c4f678.png">

alejandromumo:
After discussion with @jennur , we did not add anything else to the community item list in the deposit form.

The attributes not being shown are: 

- URL : it could be missleading as we already have a link next to the community name (`View community`). 
- Organizations: it could get cluttered.

Nonetheless if there is a strong argument on why any of the above should be included, I can add it.

PR #733, opened 2022-06-28
community: added community visibility labels to header.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-communities/issues/732

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
@@ -10,7 +10,7 @@
 {%- from "invenio_theme/macros/truncate.html" import truncate_text %}
 
 <div class="ui container fluid page-subheader-outer with-submenu rel-pt-2 ml-0-mobile mr-0-mobile">
-  <div class="ui container relaxed grid page-subheader">
+  <div class="ui container relaxed grid page-subheader mr-0-mobile ml-0-mobile">
     <div class="row pb-0">
       <div class="sixteen wide mobile sixteen wide tablet thirteen wide computer column">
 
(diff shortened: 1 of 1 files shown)


jennur:
Maybe this could go in another issue, but about the cramped space on mobile... we recently introduced `mr-0-mobile` and `ml-0-mobile` which could go on to the `.page-subheader`-grid. It would free up some space for the content (we have a lot of spacing to the edge of the screen on mobile, due to paddings and margins on nested containers).

Still, I agree with @jrcastro2 that maybe the label could go on top on mobile? 

On desktop, maybe we could just force the width of the label to make sure it doesn't break, and align it on top with the title, like this?:
Screenshot 2022-06-29 at 15 22 15

What do you think?

invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -19,7 +19,17 @@
         </div>
 
         <div>
-          <h2 class="ui header">{{ community.metadata.title }}</h2>
+          <div class="flex align-items-center rel-mb-1">
+            <h2 class="ui header mb-0">{{ community.metadata.title }}</h2>
+            {% if community.access.visibility == 'restricted' %}
+              <div class="rel-ml-1 align-self-baseline">
```
jennur, line 25:
```suggestion
              <div class="rel-ml-1">
```
I think this is actually not needed if you remove the `align-items-center` from the `flex`-element? Also, I think the empty `<div>` around the `.flex`-element can be removed. Correct me if I'm wrong!

alejandromumo:
> If a community has a long name, the label doesn't look as nice, see image:
Maybe setting `width: max-content` to the label could fix it (it's just an idea, probably not the best)


When the title has large words, the whole page breaks in mobile. I opened this [issue](https://github.com/inveniosoftware/invenio-app-rdm/issues/1770)

Changes:

- Moved the label to be inline with the title, following @jennur suggestion. 
- Added `width: max-content` for `access-status` labels, tested it in other views (e.g. deposit form) and seems to be working fine. Nonetheless, if it might break something please let me know.
- Following @kpsherva suggestion, removed the access status label for `public`.  Nonetheless I think we have this behavior elsewhere (see: https://github.com/inveniosoftware/invenio-communities/blob/5ef4cb6048f39bd07d482d3eb787e923565c5b4c/invenio_communities/assets/semantic-ui/js/invenio_communities/details_search/components.js#L83-L88)
- Also added `mr-0-mobile` and `ml-0-mobile` for the subheader page as @jennur suggested.  

Result:

Mobile 

Screenshot at Jun 30 10-43-46

Desktop

Screenshot at Jun 30 10-52-34

kpsherva (approved):

PR #738, opened 2022-06-29
settings: added logo size quota.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1716

- Added maximum size allowed for community logos.
  - By default it is set to 1 MB, can be modified per instance. 
  - The magic number 1MB was decided based on a search about image standards for web apps. For logos the recommended size was 500 KB, thus an extra 500KB were given. 

Note: this change should be documented for the instance configuration.


invenio_communities/communities/services/config.py
```diff
@@ -111,6 +111,8 @@ class CommunityServiceConfig(RecordServiceConfig):
         "{+api}/communities/{community_id}/requests{?args*}"
     )
 
+    community_logo_size_default_quota = 10**6
```
alejandromumo, line 114: Does it make sense to have a default value? I put it here to avoid issues if the app config is missing.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -168,7 +168,11 @@ const LogoUploader = ({ community, defaultLogo, hasLogo, onError }) => {
             <Icon name="upload" />
             {i18next.t("Upload new picture")}
           </Button>
-
+          <label className="helptext">
+            {i18next.t('File must be smaller than {{limit}} MB.', {
+              limit: Math.ceil(logoQuota / 10**6)
```
alejandromumo, line 173: Is there a suggestion for converting this value on a more centralized way? I noticed that we have a function `humanReadableBytes` https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/FileUploader/utils.js. but could not import it and use it here.

invenio_communities/errors.py
```diff
@@ -26,3 +27,18 @@ def __init__(self, query_arguments):
...
+
+class LogoSizeLimitError(CommunityError):
+    """The provided logo size exceeds limit."""
+
+    def __init__(self, limit, file_size):
+        """Initialise error."""
+        super().__init__(
+            _(
+                "Logo size limit exceeded. Limit: {limit} MB Given: {file_size} MB".format(
+                    limit=ceil(limit/10**6),
```
alejandromumo, line 40: Same question as before. This conversion is being done in many places, perhaps there is a better way to deal with this? What do you think?

invenio_communities/views/communities.py
```diff
@@ -118,10 +118,19 @@ def communities_settings(pid_value, community, community_ui):
     except FileNotFoundError:
         logo = False
 
+    logo_size_limit = (
+        current_communities.service.config.community_logo_size_default_quota
+    )
+    if "APP_RDM_COMMUNITY_LOGO_QUOTA" in current_app.config:
```
ntarocco, line 124: This is not ideal because `invenio-communities` should not know about `invenio-app-rdm`. Can we, instead, move the config here?

I would also align the config name with `invenio-files-rest` for consistency, see [here](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/config.py#L48). WDYT?
E.g. `COMMUNITIES_LOGO_MAX_FILE_SIZE`.

invenio_communities/templates/semantic-ui/invenio_communities/details/settings/profile.html
```diff
@@ -24,6 +24,7 @@
     data-community='{{ community | tojson}}'
     data-has-logo='{{ has_logo | tojson }}'
     data-types='{{ types | tojson }}'
+    data-logo-quota='{{ logo_quota | tojson }}'
```
ntarocco, line 27:
```suggestion
    data-logo-max-size='{{ logo_max_size| tojson }}'
```

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
ntarocco: It is just missing to be exported?

invenio_communities/views/communities.py
alejandromumo: My rationale was whether this quota size could be manageable by instance. If we move the config to the service, how could users change it? I am asking it because I am not 100% familiarised on how we handle this kind of configs in invenio RDM. 

Regarding `invenio-communities` not knowning about `app-rdm`, I think we're using `current_app.config` in many other places in communities. What's the criteria there?

I agree with the naming 👍

invenio_communities/communities/services/service.py
```diff
@@ -205,6 +209,15 @@ def update_logo(self, identity, id_, stream, content_length=None, uow=None):
         record = self.record_cls.pid.resolve(id_)
         self.require_permission(identity, "update", record=record)
 
+        logo_size_limit = self.config.community_logo_size_default_quota
+        if "APP_RDM_COMMUNITY_LOGO_QUOTA" in current_app.config:
```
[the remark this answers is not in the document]
alejandromumo: The idea here was to guarantee that at least a default value is provided by the service itself, in case `APP_RDM` does not set the config.

I think `in` operator does not throw an exception when the key is not set. Initially I had:

```python
if current_app.config["APP_RDM_COMMUNTY_LOGO_QUOTA"]:
    pass
```

and it was failing during tests with a `KeyError` exception. Using `x in dict` does not throw an exception and instead returns `false`.

I agree that the validation is indeed neded ( `is_digit()` and `>0`), just in case the config is badly set by the user.

invenio_communities/communities/services/config.py
alejandromumo: I did not understand the suggestion sorry. 

Should I move this variable outside the config file?

invenio_communities/views/communities.py
alejandromumo: About `maxStorage` I agree that it can be removed. I took other config as an example but I think it does not apply to this case. Thanks for the catch! 👍 

About tests, I will add them right away.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/index.js
```diff
@@ -168,7 +168,10 @@ const LogoUploader = ({ community, defaultLogo, hasLogo, onError }) => {
             <Icon name="upload" />
             {i18next.t("Upload new picture")}
           </Button>
-
+          <label className="helptext">
+            {i18next.t('File must be smaller than ')}
+            <>{humanReadableBytes(logoMaxSize, true)}</>
```
Pineirin, line 173: Do you need the <>?

Pineirin (approved):

kpsherva (approved):

ntarocco (approved):
LGTM! Good work!
Thank you for adding the test!

PR #751, opened 2022-07-05
header: add onerror inline js to script

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1616

- Adds onerror logic to function

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
@@ -15,7 +15,7 @@
       <div class="sixteen wide mobile sixteen wide tablet thirteen wide computer column">
 
         <div class="ui rounded image community-image mt-5 rel-mr-2 left-floated">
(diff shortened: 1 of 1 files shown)


invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -8,14 +8,21 @@
 #}
 
 {%- from "invenio_theme/macros/truncate.html" import truncate_text %}
+<!-- Image placeholder -->
+<script type="text/javascript">
+  function imagePlaceholder(image) {
+    image.onerror=null;
+    image.src='/static/images/square-placeholder.png';
+  }
+</script>
```
Pineirin, line 17: As this function seems to be repeated, could we maybe create a macro with it?

jennur: Yes, see latest comment! :)

alejandromumo (approved):

kpsherva (approved):

PR #757, opened 2022-07-08
header: added class to access-statust div to set its width to fit lab…

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1770

- Truncate community listing cards name to 30 characters, down from 50.
- Enclosed `access-status` label and icon with a new class to enforce the `min-with`. See comments for more on this.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/ComputerTabletCommunitiesItem.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/ComputerTabletCommunitiesItem.js
@@ -20,7 +20,7 @@ export const ComputerTabletCommunitiesItem = ({ result }) => {
     : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={result.id} className="computer tablet only flex">
+    <Item key={result.id} className="computer tablet only flex community-item">
       <Image
         wrapped
         src={result.links.logo}
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/MobileCommunitiesItem.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/MobileCommunitiesItem.js
@@ -20,7 +20,7 @@ export const MobileCommunitiesItem = ({ result }) => {
     : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={result.id} className="mobile only">
+    <Item key={result.id} className="mobile only community-item">
       <Item.Content className="centered">
         <Item.Extra className="user-communities">
           {community_type && (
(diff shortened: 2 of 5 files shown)


invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -21,7 +21,7 @@
           <div class="flex rel-mb-1">
             <h2 class="ui header mb-0">{{ community.metadata.title }}</h2>
             {% if community.access.visibility == 'restricted' %}
-              <div class="rel-ml-1">
+              <div class="rel-ml-1 access-status-tag">
```
alejandromumo, line 24: I am not sure whether this is as good approach. 

The problem is that the `access-status` label has two elements: `span` and `icon`. In some cases, they can break line if there is no space for the label to render. To fix this, I set the property `min-width: fit-content`  to enforce that the whole div is always capable of containing both elements.

We can enforce this behavior in every `access-status` label that has an icon, but I think it only breaks in the community header because we're using flex.

Please feel free to suggest an alternative solution.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/frontpage.js
```diff
@@ -62,7 +62,7 @@ class CommunityCard extends Component {
         />
         <Card.Content>
           <Card.Header>
-            {_truncate(community.metadata.title, { length: 50 })}
+            {_truncate(community.metadata.title, { length: 30 })}
```
alejandromumo, line 65: With length 50, big names were being rendered  outside of the card. In mobile 30 seems to be good enough as well.

jennur:
## Updates

Moved the label to above the title on mobile, as the text content got very smashed.

## Screenshots

### Mobile
Screenshot 2022-09-26 at 17 04 39
e

#### Tablet
Screenshot 2022-09-26 at 17 04 48

### Desktop

Screenshot 2022-09-26 at 17 11 06

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/ComputerTabletCommunitiesItem.js
```diff
@@ -20,7 +20,7 @@ export const ComputerTabletCommunitiesItem = ({ result }) => {
     : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={result.id} className="computer tablet only flex">
+    <Item key={result.id} className="computer tablet only flex community-item">
```
alejandromumo, line 23: This seems to be implemented in`app-rdm` and `invenio-communities`
However, I fixed it in both places.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communities_items/MobileCommunitiesItem.js
```diff
@@ -20,7 +20,7 @@ export const MobileCommunitiesItem = ({ result }) => {
     : i18next.t("Restricted");
   const visibilityIcon = isPublic ? undefined : "ban";
   return (
-    <Item key={result.id} className="mobile only">
+    <Item key={result.id} className="mobile only community-item">
```
alejandromumo, line 23: This seems to be implemented in`app-rdm` and `invenio-communities`
However, I fixed it in both places.

kpsherva (approved):

PR #762, opened 2022-07-08
Communities search: fix mobile pagination [DEPENDS ON https://github.com/inveniosoftware/invenio-search-ui/pull/150]

Pineirin:
Screenshot from 2022-07-19 09-57-38
Screenshot from 2022-07-19 09-57-48

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/community/search.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/community/search.js
@@ -10,6 +10,7 @@ import { createSearchAppInit } from "@js/invenio_search_ui";
 import {
   SearchAppFacets,
   SearchAppResultsPane,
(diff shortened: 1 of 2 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/community/search.js
```diff
@@ -124,14 +128,33 @@ export const CommunitiesResults = ({
               }}
             />
           </Grid.Column>
-          <Grid.Column textAlign="right" width={4}>
+          <Grid.Column className="mobile only" width={16} textAlign="center">
```
NRodriguezcuellar, line 131: maybe it has been discussed already, could it be an idea to add a `common` folder and add it in there so it can be reused? I think they are identical.

kpsherva (approved):

Issue #794, opened 2022-09-14
custom fields: add config validation to `init` cli


ppanero:

ppanero:
missing app-rdm pr merged

PR #795, opened 2022-09-14
cli: add custom fields validation before creation

ppanero:
- closes https://github.com/inveniosoftware/invenio-communities/issues/794


invenio_communities/config.py
```diff
@@ -202,6 +202,22 @@
 """Community logo size quota, in bytes."""
 
 
+COMMUNITIES_NAMESPACES = {}
```
ppanero, line 205: this should be overwritten in rdm-records

`COMMUNITIES_NAMESPACES = RDM_NAMESPACES`

PR #807, opened 2022-09-19
service: add system_created field to OAI set

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1037

### DIFF
--- a/invenio_communities/communities/services/components.py
+++ b/invenio_communities/communities/services/components.py
@@ -142,6 +142,7 @@ class OAISetComponent(ServiceComponent):
 
         community_set = OAISet()
         community_set.name = community_title
+        community_set.system_created = True
         community_set.description = self._create_set_description(community_title)
         community_set.spec = self._create_set_spec(community_slug)
         community_set.search_pattern = "parent.communities.ids:{id}".format(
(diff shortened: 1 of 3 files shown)


invenio_communities/communities/services/components.py
```diff
@@ -142,6 +142,7 @@ def _create_set_from_community(self, record):
 
         community_set = OAISet()
         community_set.name = community_title
+        community_set.system_created = True
```
jrcastro2, line 145: IMHO I think we should create the OAI set by using the service and not by persisting the object directly in the database

kpsherva (approved):

invenio_communities/communities/services/components.py
ntarocco: Yes indeed. The service might do extra things. The service should probably get this extra flag `system_created` `True|False` and persist it.

kpsherva: it was discussed with @lnielsen, we can keep the direct usage of data layer as it is in this case, since there is no need for extra things that service provides, and it can be bypassed

slint (approved):

PR #817, opened 2022-10-14
assets: upgrade yup

kpsherva:


alejandromumo (approved):
Checked the [changelog](https://github.com/jquense/yup/blob/master/CHANGELOG.md) for breaking changes and it seems  OK for us to upgrade `yup`. 

These two versions had potential breaking changes but I don't think we're affected ([0.31.0](https://github.com/jquense/yup/blob/master/CHANGELOG.md#breaking-changes-4) or [0.32.0](https://github.com/jquense/yup/blob/master/CHANGELOG.md#breaking-changes-3))

ppanero (commented):
could it be "centrally managed"? If we bump the version of `react-invenio-forms` once this has been merged <https://github.com/inveniosoftware/react-invenio-forms/pull/167> then there is no reason to bump (nor add?) here?

kpsherva:
React-invenio-forms has only peer dependencies thats why this was added in other modules as well. ping @zzacharo

zzacharo (approved):

PR #828, opened 2022-10-27
Bulk indexing for members/invitations `rebuild_index`

max-moser:
Make the `service.rebuild_index` method use bulk indexing rather than indexing each item individually.
Related to https://github.com/inveniosoftware/invenio-records-resources/pull/419.

### DIFF
--- a/invenio_communities/members/services/service.py
+++ b/invenio_communities/members/services/service.py
@@ -29,14 +29,15 @@ from invenio_records_resources.services.uow import (
 from invenio_requests import current_events_service, current_requests_service
 from invenio_requests.customizations.event_types import CommentEventType
 from invenio_search.engine import dsl
+from kombu import Queue
 from marshmallow import ValidationError
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
-
-from invenio_communities.members.records.api import ArchivedInvitation
+from werkzeug.local import LocalProxy
 
 from ...proxies import current_roles
 from ..errors import AlreadyMemberError, InvalidMemberError
+from ..records.api import ArchivedInvitation
 from .request import CommunityInvitation
 from .schemas import (
     AddBulkSchema,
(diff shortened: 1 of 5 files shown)


invenio_communities/views/api.py
```diff
@@ -25,9 +25,15 @@ def init(state):
 
     # services
     rr_ext.registry.register(ext.service)
+    rr_ext.registry.register(ext.service.members, service_id="community-members")
```
ppanero, line 28:
```suggestion
    rr_ext.registry.register(ext.service.members, service_id="members")
```

to be consistent with indexer naming?

invenio_communities/members/services/service.py
```diff
@@ -650,16 +665,12 @@ def rebuild_index(self, identity, uow=None):
...
         """
-        for rec_meta in self.record_cls.model_cls.query.all():
-            rec = self.record_cls(rec_meta.data, model=rec_meta)
-
-            if not rec.is_deleted:
-                self.indexer.index(rec)
-
-        for rec_meta in ArchivedInvitation.model_cls.query.all():
-            rec = ArchivedInvitation(rec_meta.data, model=rec_meta)
+        members = self.record_cls.model_cls.query.filter_by(is_deleted=False).all()
```
ppanero, line 668: same old question about removing the `if`. Feel free to answer to it in only one of the PRs 😅

invenio_communities/views/api.py
max-moser: will amend the commit with the change on monday

max-moser: updated, and explicitly set the `service_id` for the communities service too, for consistency within this function

ppanero (approved):
LGTM

# ⚠️ haven't tested yet.

invenio_communities/views/api.py
```diff
@@ -24,10 +24,16 @@ def init(state):
     ext = app.extensions["invenio-communities"]
 
     # services
-    rr_ext.registry.register(ext.service)
+    rr_ext.registry.register(ext.service, service_id="communities")
```
ppanero, line 27: very minor: would it make sense to do `ext.service.id` (see [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/base/service.py#L76)). Although this is something we have not done in any other registration (maybe this is the point when we start fixing it hehe)

PR #839, opened 2022-11-16
tests: add storage class list

zzacharo:


zzacharo:
integrated the solution in https://github.com/inveniosoftware/invenio-communities/pull/828

PR #858, opened 2022-12-21
communities: fix destructuring statement

kpsherva:
* the destructuring was causing actionParam evaluated as undefined

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMemberBulkActions.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMemberBulkActions.js
@@ -156,13 +156,13 @@ class ManagerMemberBulkActionsCmp extends Component {
     const currentActionRender =
       this.currentAction?.renderOnActive && this.currentAction.renderOnActive();
 
-    const { text: currentActionText, actionParam } = this.currentAction;
+    const currentActionText = this.currentAction?.text;
 
     const actionDisabled =
       loading ||
       selectedCount === 0 ||
       // eslint-disable-next-line react/destructuring-assignment
-      !this.state[actionParam] === undefined;
+      !this.state[this.currentAction.actionParam] === undefined;
 
     return (
       <>


jrcastro2 (approved):
LGTM, although I find it weird it throws such a warning ... maybe we are missing some rules in eslint and we should consider enforcing destructuring, if it makes sense, I guess we would need to add the [prefer-destructuring](https://eslint.org/docs/latest/rules/prefer-destructuring) rule to `eslint-config-invenio`, WDYT?

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMemberBulkActions.js
```diff
@@ -156,13 +156,13 @@ class ManagerMemberBulkActionsCmp extends Component {
...
 
-    const { text: currentActionText, actionParam } = this.currentAction;
+    const currentActionText = this.currentAction?.text;
 
     const actionDisabled =
       loading ||
       selectedCount === 0 ||
       // eslint-disable-next-line react/destructuring-assignment
-      !this.state[actionParam] === undefined;
+      !this.state[this.currentAction.actionParam] === undefined;
```
Pineirin, line 165: I had this issue when fixing the linting warnings.
I managed to keep the destructuring by checking the variable before using it, so maybe it could be an option.

Pineirin (approved):

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMemberBulkActions.js
kpsherva: this is not a variable but a getter function, that is why there is a problem with destructuring, see the answer here on why this is happening
https://stackoverflow.com/questions/69580147/destructuring-losing-getter-definition

kpsherva:
> LGTM, although I find it weird it throws such a warning ... maybe we are missing some rules in eslint and we should consider enforcing destructuring, if it makes sense, I guess we would need to add the [prefer-destructuring](https://eslint.org/docs/latest/rules/prefer-destructuring) rule to `eslint-config-invenio`, WDYT?

see my answer here: https://github.com/inveniosoftware/invenio-communities/pull/858#discussion_r1054232959

I would not relax the rules because of the usage of getters

PR #859, opened 2023-01-16
requests: update label size to tiny

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1801

Set to same size as in the uploads page (see [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/4519b4863faddf7b92f1462417937dac20d60431/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js#L51))
Screenshot from 2023-01-16 15-11-39

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/requests/index.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/requests/index.js
@@ -76,24 +76,24 @@ const RequestCancelModalTriggerWithConfig = parametrize(RequestCancelModalTrigge
 });
 
 const CommunitySubmission = () => (
-  <LabelTypeSubmission className="primary" size="small" />
+  <LabelTypeSubmission className="primary" size="tiny" />
 );
 
 const CommunityInvitation = () => (
-  <LabelTypeInvitation className="primary" size="small" />
+  <LabelTypeInvitation className="primary" size="tiny" />
 );
 
-const Submitted = () => <LabelStatusSubmit className="primary" size="small" />;
+const Submitted = () => <LabelStatusSubmit className="primary" size="tiny" />;
 
-const Deleted = () => <LabelStatusDelete className="negative" size="small" />;
+const Deleted = () => <LabelStatusDelete className="negative" size="tiny" />;
 
-const Accepted = () => <LabelStatusAccept className="positive" size="small" />;
+const Accepted = () => <LabelStatusAccept className="positive" size="tiny" />;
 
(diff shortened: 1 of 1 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/requests/index.js
```diff
@@ -76,24 +76,24 @@ const RequestCancelModalTriggerWithConfig = parametrize(RequestCancelModalTrigge
...
+  <LabelTypeSubmission className="primary" size="tiny" />
 );
 
 const CommunityInvitation = () => (
-  <LabelTypeInvitation className="primary" size="small" />
+  <LabelTypeInvitation className="primary" size="tiny" />
 );
 
-const Submitted = () => <LabelStatusSubmit className="primary" size="small" />;
+const Submitted = () => <LabelStatusSubmit className="primary" size="tiny" />;
```
kpsherva, line 86: I would be a bit worried that it won't be readable anymore

jrcastro2: IMHO it's readable. It would be the same size as we have for the uploads page. In that same view we have some smaller text which I think is readable. However, if we want to highlight these labels a bit more, then I think we should increase the size for **all** the labels and find an alternative solution to center the icon on the left. 

Here an example of the uploads page with the **tiny** size .
Screenshot from 2023-01-17 09-43-00

kpsherva: I think you might be biased by the font changes in your PC/browser. IMO it is not readable, at least not on my screen.
Sacrificing readability and user experience, because of icon misalignment is not very well justified.

jrcastro2: Sorry I've not explained my self well. 
I just tried to align the icons on the entire website, as I mentioned in my previos comment the upload page is using the size tiny to display the tags while the requests page is using the size small. Since the alignment for the tiny labels is correct and we were using them already I thought it would make sense to harmonize them across the entire site. If you think that this size is not readable we should revise and update the uploads page because is currently using it (see [here an example](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/user_dashboard/uploads_items/ComputerTabletUploadsItem.js#L52) of it)

Issue #867, opened 2023-01-20
Design new details sections on community details page


jennur:
Design and implement display of the data as new community tab(s) in InvenioRDM

Part 2 of https://github.com/inveniosoftware/invenio-communities/issues/861

jennur:
## Suggested layout

- Curation policy goes in popup ([SUI popup](https://react.semantic-ui.com/modules/popup/)) in the community header
    - Should only be opened/closed by click
    - Need to make sure it is accessible
    - Props:
        - `trigger={<span><Icon name='info' circular /> Curation policy</span>}`
        - `wide`
        - `on="click"`
        - `position="bottom right"` on computer, `position="bottom left"` on tablet/mobile
 - "About this community" is moved from the header to a separate tab in the tab menu.
 - "Want your upload to appear in this community" instructions is present on every community's about page.

### Desktop
![Image](https://user-images.githubusercontent.com/21052053/214303349-8ffb6d6c-9e36-49fe-a7e5-2ef5f64a3522.png)
![Image](https://user-images.githubusercontent.com/21052053/214303350-5d55b736-cf5e-4f3b-a26a-6fcf1df6202e.png)

### Tablet
![Image](https://user-images.githubusercontent.com/21052053/214303956-c364ceee-d660-4eaa-953b-23ae0d7ecb35.png)

### Mobile
![Image](https://user-images.githubusercontent.com/21052053/214304801-e1853b1d-98d1-419a-bfb8-c38c58b38ca7.png)

jennur:
## Curation policy in popup
- Curation policy goes in popup ([SUI popup](https://react.semantic-ui.com/modules/popup/)) in the community header
    - Should only be opened/closed by click
    - Need to make sure it is accessible
    - Props:
        - `trigger={<span><Icon name='info' circular /> Curation policy</span>}`
        - `wide`
        - `on="click"`
        - `position="bottom right"` on computer, `position="bottom left"` on tablet/mobile
 - Curation details (curator, created date, etc) is displayed below the custom about-text.
    - **Question:** Should we keep the curation policy here as well? (Maybe it's good to display it both places? LMK and I will add it to the mockup so that it reflects the popup content.)

![Image](https://user-images.githubusercontent.com/21052053/214303350-5d55b736-cf5e-4f3b-a26a-6fcf1df6202e.png)
[...]

PR #884, opened 2023-02-03
community details: add about and curation policy tab

jennur:
Closes https://github.com/inveniosoftware/invenio-communities/issues/867

Needs https://github.com/inveniosoftware/invenio-app-rdm/pull/2057

## Changes
- Moved the about section from the header to new tab
- Added  new tab for curation policy
- Added new tab on the settings page for curation policy
- Added TODO comment for updating the upload-instructions text

### Settings page
-  I renamed the "Description" field in the settings page to "Summary", as we have the "Page description" below. This will show up above the "page description" on the about page (see above divider in the about page - next screenshot). Let me know if this should be renamed, or solved differently?
Screenshot 2023-02-06 at 14 46 28

### About page
Screenshot 2023-02-03 at 17 18 01

### Curation policy page
Screenshot 2023-02-03 at 17 15 11

### Curation policy settings page
Screenshot 2023-02-03 at 16 43 37

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/about/index.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/about/index.html
@@ -1,0 +1,62 @@
+{# -*- coding: utf-8 -*-
+
+  This file is part of Invenio.
+  Copyright (C) 2023 CERN.
+
+  Invenio is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% extends "invenio_communities/details/base.html" %}
+{% set active_community_header_menu_item= 'about' %}
+
+{%- block page_body %}
+  {{ super() }}
+  <div class="ui text container rel-mt-2">
+    <h2 class="ui header rel-mt-2">{{ _("About this community") }}</h2>
+    <p class="text-muted">{{ _("Created") }} {{ community.created|invenio_format_datetime }}</p>
+    <p>{{ community.metadata.description | safe }}</p>
+
+    {% if community.ui.funding|length %}
+      <h3 class="ui header">{{ _("Awards") }}</h3>
+      <dl class="ui list">
+
+        {% for funding in community.ui.funding %}
+          <div class="item rel-mb-1">
+            {% if funding.award %}
+              <dd class="header">
+                {{ funding.award.title_l10n }}
+                <label class="ui basic small label">
+                  {{funding.award.number}}
+                </label>
+
+                {% if funding.award.identifiers|length and funding.award.identifiers[0].scheme == "url" %}
+                  <a
+                    class="ui transparent icon button"
+                    href="{{ funding.award.identifiers[0].identifier }}"
+                    aria-label="{{ _('Visit external website') }}"
+                    title="{{ _('Opens in new tab') }}"
+                  >
+                    <i class="external primary icon" aria-hidden="true"></i>
+                  </a>
+                {% endif %}
+              </dd>
+            {% endif %}
+
+            {% if funding.funder %}
+              <dt class="text-muted">
+                {{ funding.funder.name }}
+              </dt>
+            {% endif %}
+          </div>
+        {% endfor %}
+
+      </dl>
+    {% endif %}
+  </div>
+
+  {% include "invenio_communities/details/upload-instructions.html" %}
+
+
+
+{%- endblock page_body -%}
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/settings/curation_policy.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/settings/curation_policy.html
@@ -1,0 +1,27 @@
+{# -*- coding: utf-8 -*-
+
+  This file is part of Invenio.
+  Copyright (C) 2023 CERN.
+
+  Invenio is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% extends "invenio_communities/details/settings/base.html" %}
+
+{%- block javascript %}
+{{ super() }}
+{{ webpack['invenio-communities-curation-policy.js'] }}
+{%- endblock javascript %}
+
+{% set active_settings_menu_item = 'curation_policy' %}
+{% set active_community_header_menu_item = 'settings' %}
+
+{%- block settings_body %}
+<div class="sixteen wide mobile sixteen wide tablet thirteen wide computer column">
+  <div
+    id="app"
+    data-community='{{ community | tojson }}'>
+  </div>
+</div>
+{%- endblock settings_body %}
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/upload-instructions.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/upload-instructions.html
@@ -1,0 +1,13 @@
+<div class="ui container rel-mt-2">
+    <div class="ui divider" aria-hidden="true"></div>
+    <div class="ui text container rel-mt-2">
+
+    <h2 class="ui small header">Want your upload to appear in this community?</h2>
+      <ul>
+        <!-- TODO: This text content needs to be updated (Copied from Zenodo) -->
+        <li>Click the "New upload" button above to upload a record directly to this community.</li>
+        <li>The community curator will then be notified to either accept or reject your upload (see the community <a href="{{ url_for('invenio_communities.communities_curation_policy', pid_value=community.slug) }}">curation policy</a>).</li>
+        <li>If your upload is rejected by the curator, it will still be available, just not in this community.</li>
+      </ul>
+    </div>
+  </div>
--- a/invenio_communities/views/communities.py
+++ b/invenio_communities/views/communities.py
@@ -217,6 +217,20 @@ def communities_settings_privileges(pid_value, community, community_ui):
 
 
 @pass_community(serialize=True)
+def communities_settings_curation_policy(pid_value, community, community_ui):
+    """Community settings/curation-policy page."""
+    permissions = community.has_permissions_to(["update", "read", "search_requests"])
+    if not permissions["can_update"]:
+        raise PermissionDeniedError()
+
+    return render_template(
+        "invenio_communities/details/settings/curation_policy.html",
+        community=community_ui,
+        permissions=permissions,
+    )
+
+
+@pass_community(serialize=True)
 def members(pid_value, community, community_ui):
     """Community members page."""
     permissions = community.has_permissions_to(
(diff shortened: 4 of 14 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
```diff
@@ -0,0 +1,146 @@
...
+            <Grid>
+              <Grid.Row className="pt-10 pb-0">
+                <Grid.Column mobile={16} tablet={16} computer={12} className="rel-pb-2">
+                  <RichInputField
+                    fieldPath="metadata.curation_policy"
+                    label={
+                      <FieldLabel
+                        htmlFor="metadata.curation_policy"
+                        icon="pencil"
+                        label="Curation policy"
```
jrcastro2, line 109: Minor: Missing internalization.

invenio_communities/templates/semantic-ui/invenio_communities/details/about/index.html
```diff
@@ -0,0 +1,64 @@
...
+              </dt>
+            {% endif %}
+          </div>
+        {% endfor %}
+
+      </dl>
+    {% endif %}
+  </div>
+
+  {{ upload_instructions() }}
```
jrcastro2, line 61: Very very minor: Sorry for being picky ... Ideally macros are useful when we need to pass dynamic parameters to it, if it's static content we can simply use include.

invenio_communities/views/communities.py
```diff
@@ -216,6 +216,22 @@ def communities_settings_privileges(pid_value, community, community_ui):
     )
 
 
+@pass_community(serialize=True)
+def communities_settings_curation_policy(pid_value, community, community_ui):
+    """Community settings/curation-policy page."""
+    permissions = community.has_permissions_to(
+        ["update", "read", "search_requests", "search_invites"]
```
jrcastro2, line 223: Minor: Do we need to check for all these permissions? wouldn't be enough the "update"? 

Same question for the functions below.

invenio_communities/views/communities.py
```diff
@@ -216,6 +216,22 @@ def communities_settings_privileges(pid_value, community, community_ui):
...
+    )
+    if not permissions["can_update"]:
+        raise PermissionDeniedError()
+
+    return render_template(
+        "invenio_communities/details/settings/curation_policy.html",
+        community=community_ui,
+        permissions=permissions,
```
jrcastro2, line 231: Minor: Do we need to pass the permissions? I've checked and looks like it's not really used (unless I am mistaken).

Same comment for the functions below.

jrcastro2 (commented):
Looks nice!
I just left a few minor comments and questions.

invenio_communities/templates/semantic-ui/invenio_communities/details/about/index.html
jennur: You're right! Will fix. Thanks!

invenio_communities/views/communities.py
jennur: Will test! 🙏

invenio_communities/views/communities.py
jennur: They are used in the submenu in the header: 
https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html#L133

ntarocco:
Can you please rebase?

invenio_communities/views/ui.py
```diff
@@ -148,4 +182,16 @@ def register_menus():
     # Register context processor
     blueprint.app_context_processor(search_app_context)
 
+    # Template filters
+    @blueprint.app_template_filter()
+    def format_datetime(value, format="medium"):
```
ntarocco, line 187: Shouldn't be the date/time formatted by current `locale`?

jennur: You're right, thanks!

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
```diff
@@ -0,0 +1,146 @@
...
+        this.setGlobalError(message);
+      }
+
+      if (errors) {
+        errors.forEach(({ field, messages }) => setFieldError(field, messages[0]));
+      }
+    }
+    this.setIsSavedState(true);
+
+    const savedStateTimer = setTimeout(() => {
```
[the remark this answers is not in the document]
jennur: We don't need it for the functionality to work, the user just won't know if the save was successful or not (i.e. no feedback. which is not great). I don't know if we could maybe add an `onChange` event to the `RichInputField` component in react-invenio-forms and then solve it like we do in "Privileges"?

invenio_communities/templates/semantic-ui/invenio_communities/details/upload-instructions.html
```diff
@@ -0,0 +1,14 @@
+<div class="ui container rel-mt-2">
+    <div class="ui divider" aria-hidden="true"></div>
+    <div class="ui text container rel-mt-2">
+
+    <h2 class="ui small header">Want your upload to appear in this community?</h2>
+      <ul>
+        <!-- TODO: This text content needs to be updated (Copied from Zenodo) -->
```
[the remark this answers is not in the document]
jennur: I added it there because the text is not completely up to date (it's copied from Zenodo). But as the workflow is not completely decided yet, I guess we don't know what will be the instructions?

zzacharo:
>  renamed the "Description" field in the settings page to "Summary", as we have the "Page description" below. This will show up above the "page description" on the about page

what about "Community description" with a help text that this text will be shown on the About page? I would probably keep the description name untouched and indicating that this is a short description of the community. If the user wants to provide a detailed text I would encourage them to use the About field....TLDR: I would take advantage of the help texts to guide the user how to fill the fields...

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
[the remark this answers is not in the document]
kpsherva: I am not sure I understand the reason for using timer, we have the response from backend, which we can use to communicate the state to the user, alongside with the refreshed values. Unless I am missing something

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
```diff
@@ -0,0 +1,146 @@
...
+import React, { Component } from "react";
+import ReactDOM from "react-dom";
+import { FieldLabel, RichInputField, CKEditorConfig } from "react-invenio-forms";
+import { Button, Form, Grid, Icon, Message } from "semantic-ui-react";
+import { CommunityApi } from "../api";
+import { communityErrorSerializer } from "../api/serializers";
+import PropTypes from "prop-types";
+import * as Yup from "yup";
+
+const COMMUNITY_VALIDATION_SCHEMA = Yup.object({
```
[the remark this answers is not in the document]
slint: I think for small forms, it's actually nice to have client-side validation and quick feedback (vs. having to always submit to the backend to get the validation errors). IIRC Formik also has baked-in support for Yup, and that's why we use it.

The deposit form is a beast on its own, so I think the "slow" backend-based feedback is fine (and also allows the form to be "lighter").

So I'm pro-Yup, but would be fine also if we decided to skip it.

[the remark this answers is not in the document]
ppanero: I think the more centalized validation is better, we currently have:
- Marshmallow in the backend
- Very light on formik (?)

We would add yup on the frontend. I'm not against this if it improves the UX, however we should document in some _best practices_ section a clear cut of when is it used and when is it not. To avoid having inconsistencies.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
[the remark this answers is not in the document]
jennur: I will look more into it...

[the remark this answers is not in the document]
jennur: We can use the state from the backend to show a saved state for the button, but the problem is that it never resets. I.e. it will show "saved" even when the user changes the text again. In the "privileges" form, the saved-state is reset on `onChange` on the fields, but there is no `onChange` event on the `RichInputField`, so there is no way to reset the state in this form. That's why I put the timer as a hack just to give some feedback to the user. On the "profile" page, the entire page reloads, so I guess we can do that instead for this page as well.

jennur:
## Updates

- Removed the saved-state timer
- Added a page reload after save (similar to the profile settings page)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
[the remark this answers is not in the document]
ntarocco: Yup is already used in `react-invenio-deposit`, `react-invenio-forms` and `invenio-communities` at least, see [here](https://github.com/search?q=org%3Ainveniosoftware+yup+extension%3A.js&type=Code&ref=advsearch&l=&l=).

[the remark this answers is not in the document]
kpsherva: I have seen it, but I think it was maybe introduced without discussion. Maybe it would be better to keep validation consistent, as Pablo says, if not documented it might diverge even more

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/curation-policy.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+CommunityCurationPolicyForm.propTypes = {
+  community: PropTypes.object.isRequired,
+};
+
+const domContainer = document.getElementById("app");
+const community = JSON.parse(domContainer.dataset.community);
+
+ReactDOM.render(<CommunityCurationPolicyForm community={community} />, domContainer);
```
kpsherva, line 132: could this be moved to separate `index.js` file ? then we keep only the component class here

invenio_communities/templates/semantic-ui/invenio_communities/details/settings/curation_policy.html
```diff
@@ -0,0 +1,27 @@
...
+{%- endblock javascript %}
+
+{% set active_settings_menu_item = 'curation_policy' %}
+{% set active_community_header_menu_item = 'settings' %}
+
+{%- block settings_body %}
+<div class="sixteen wide mobile sixteen wide tablet thirteen wide computer column">
+  <div
+    id="app"
+    data-community='{{ community | tojson}}'>
```
kpsherva, line 24: nit: 
```suggestion
    data-community='{{ community | tojson }}'>
```

invenio_communities/templates/semantic-ui/invenio_communities/details/upload-instructions.html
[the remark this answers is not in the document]
kpsherva: could we at least change the "Zenodo" part, that it is mergeable ? Then we can re-iterate on the content

kpsherva (approved):

zzacharo (approved):

Issue #896, opened 2023-02-20
members: serialize read_memberships response


ppanero:
The `members_service.read_memberships()` returns a hardcoded dictionary. A proper serializations/result_item_cls should be used.

PR #897, opened 2023-02-20
members: add service method to return identity memberships

ppanero:


invenio_communities/members/services/service.py
```diff
@@ -437,6 +437,10 @@ def _members_search(
             schema=schema,
         )
 
+    def read_memberships(self, identity):
+        """Searches the memberships of a specific user/identity."""
+        return {"memberships": self.config.record_cls.get_memberships(identity)}
```
ppanero, line 442: This is not an ideal searialization. Doing it properly would require rework of the actual members cls item.
To be done separately at https://github.com/inveniosoftware/invenio-communities/issues/896

jrcastro2 (approved):

PR #901, opened 2023-02-23
wdiget: added multiple CF widget loaders.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1191

- Updated custom fields widget loading from multiple sources


invenio_communities/assets/semantic-ui/js/invenio_communities/community/new.js
```diff
@@ -155,9 +155,11 @@ class CommunityCreateForm extends Component {
                   {!_isEmpty(customFields.ui) && (
                     <CustomFields
                       config={customFields.ui}
-                      templateLoader={(widget) =>
-                        import(`@templates/custom_fields/${widget}.js`)
-                      }
+                      templateLoaders={[
+                        (widget) => import(`@templates/custom_fields/${widget}.js`),
+                        (widget) => import(`react-invenio-deposit`),
+                        (widget) => import(`react-invenio-forms`),
```
kpsherva, line 161: shouldn't we try first from the top hierarchy modules and then lastly from the instance? otherwise my gut feeling tells me that if you override something in your instance (with the same file name) then it will never be picked up - because the last import will be from react-invenio-forms.

alejandromumo: That depends on the implementation of the loader itself. We had the same discussion when implementing this. 

Right now, the loader stops ([`break`](https://github.com/inveniosoftware/react-invenio-forms/blob/0683a84db131fd6f654fafe2851229ef08cd1145/src/lib/forms/widgets/loader.js#L23)) at the first match.

kpsherva (approved):

PR #908, opened 2023-03-03
resolve deleted communities

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2063


invenio_communities/communities/resolver.py
```diff
@@ -27,15 +27,38 @@
 from .services.config import CommunityServiceConfig
 
 
+def tombstone_community(id):
+    """Return a tombstone community."""
+    return {
```
ppanero, line 32: moderate: this is hardcoding a dict/datamodel, would it be worth it to use [CommunitySchema](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/schema.py#L114), or at least dump it to make sure we comply?

invenio_communities/communities/resolver.py
ppanero, line 1: IMO, this breaks the resources-services-datalayer design. The distribution of files/python modules in invenio-communities is already quite messy and it's not respecting what `invenio-records-resources` defines (that is another topic --> https://github.com/inveniosoftware/invenio-communities/issues/922).

Nonetheless, this is adding yet more inconsistency. Resolvers are meant to be data layer access, see [invenio-records-resources](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/resolver.py) base record resolver, then it should be used by forwarding the call resource -> service -> resolver (and return back).

invenio_communities/errors.py
```diff
@@ -42,3 +42,17 @@ def __init__(self, limit, file_size):
                 )
             )
         )
+
+
+class OpenRequestsForCommunitytError(CommunityError):
+    """The provided logo size exceeds limit."""
```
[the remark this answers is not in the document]
ntarocco: Yes, the docstring is a copy/paste from another class.

invenio_communities/errors.py
```diff
@@ -42,3 +42,17 @@ def __init__(self, limit, file_size):
...
+
+
+class OpenRequestsForCommunityError(CommunityError):
+    """The provided logo size exceeds limit."""
+
+    def __init__(self, requests):
+        """Initialise error."""
+        super().__init__(
+            _(
+                "There are {requests} request(s) open for this community. Please, resolve all of them before deleting this community.".format(
```
ntarocco, line 54: With translations, can we add the singular/plural forms, instead of `There are ... request(s)`?

ntarocco (approved):

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/CommunityProfileForm.js
```diff
@@ -286,7 +287,12 @@ class CommunityProfileForm extends Component {
 
   setGlobalError = (error) => {
     const { message } = communityErrorSerializer(error);
-    this.setState({ error: message });
+    this.setState({ error: message }, () =>
+      this.globalErrorRef.current.scrollIntoView({
```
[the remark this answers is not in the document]
ntarocco: It was discussed to fix the error UI in the next PR, after mock-ups.

invenio_communities/communities/entity_resolver.py
```diff
@@ -24,18 +24,26 @@
 from ..generators import CommunityRoleNeed
 from ..proxies import current_communities, current_roles
 from .records.api import Community
+from .schema import CommunityGhostSchema
 from .services.config import CommunityServiceConfig
 
 
 def pick_fields(identity, community_dict):
     """Pick fields to return when expanding the community obj."""
+    if community_dict.get("is_ghost"):
```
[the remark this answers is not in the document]
zzacharo: Just because the ghost representation of the community doesn't have all the fields returned below. I will add a comment.

invenio_communities/errors.py
zzacharo: will create an issue for that :)

PR #925, opened 2023-03-24
search: added terms safelisting

alejandromumo:
closes https://github.com/inveniosoftware/invenio-records-resources/issues/447

### DIFF
--- a/invenio_communities/members/services/config.py
+++ b/invenio_communities/members/services/config.py
@@ -54,16 +54,20 @@ class PublicSearchOptions(SearchOptions):
             "user.profile.full_name^3",
             "user.profile.affiliations",
         ],
-        tree_transformer_factory=SearchFieldTransformer.factory(
-            mapping={
-                "affiliation": "user.profile.affiliations",
-                "affiliations": "user.profile.affiliations",
-                "full_name": "user.profile.full_name",
-                "fullname": "user.profile.full_name",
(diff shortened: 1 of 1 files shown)


alejandromumo:
Tests might fail because it depends on https://github.com/inveniosoftware/invenio-records-resources/pull/462

invenio_communities/members/services/config.py
```diff
@@ -55,6 +55,7 @@ class PublicSearchOptions(SearchOptions):
             "user.profile.affiliations",
         ],
         tree_transformer_factory=SearchFieldTransformer.factory(
+            allow_list=["user.profile.affiliations", "user.profile.full_name", "user.username"],
```
ppanero, line 58: missing `user.username`?

ppanero (commented):
is it worth having some tests?

ppanero (approved):
Discussed with Karolina IRL, @alejandromumo I'm approving this but there's a need for tests in a diff PR.

alejandromumo:
> Discussed with Karolina IRL, @alejandromumo I'm approving this but there's a need for tests in a diff PR.

Good point, I only updated tests for the query parser. And given how many things happen in this code, maybe it's a good idea to have more tests.

ppanero:
test passing locally, requires first users-resources and records-resources
Screenshot 2023-04-17 at 14 35 40

PR #927, opened 2023-03-27
resource: allow to exclude communities on search

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2076

### DIFF
--- a/invenio_communities/communities/resources/resource.py
+++ b/invenio_communities/communities/resources/resource.py
@@ -28,11 +28,16 @@ from invenio_records_resources.resources.records.resource import (
     request_view_args,
 )
 from invenio_records_resources.resources.records.utils import search_preference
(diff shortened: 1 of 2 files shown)


invenio_communities/communities/resources/resource.py
```diff
@@ -61,6 +66,28 @@ def create_url_rules(self):
             route("GET", routes["community-requests"], self.search_community_requests),
         ]
 
+    @excluded_communities
```
kpsherva, line 69: wouldn't it be easier to add excluded_communities in the request_search_args instead of creating the new decorator?

jrcastro2: Yes, we could, but that decorator is used by other resources that do not need the `excluded_communities`, therefore I thought about creating a dedicated one.

Issue #935, opened 2023-03-29
Count members and records of community


anikachurilova:
Integrate information about amount of members and records of the community for the delete communinty modal

Before:
<img width="1552" alt="Screenshot 2023-03-29 at 14 00 51" src="https://user-images.githubusercontent.com/61321254/228553426-cf9636b5-597b-41cb-b7b6-9a2b4e7cd88e.png">

After:
<img width="536" alt="Screenshot 2023-03-29 at 15 28 52" src="https://user-images.githubusercontent.com/61321254/228553494-8d31d864-11d4-4f79-be1c-46218a02e9ea.png">

kpsherva:
hints:

check the REST API of members service and see how you can dump this information in settings page

probably you can dump this information in the view corresponding to the settings page so you have them available already

PR #944, opened 2023-04-19
delete community: display metrics on deletion modal

mitsosf:
* closes https://github.com/inveniosoftware/invenio-communities/issues/935

### Description

When a user decides to delete a community the Community Deletion Modal pops up, to make sure the user is aware of the impact of their actions. Some community metrics (member and record counts) are now displayed on the Community Deletion Modal. 

### Checklist

- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
@@ -8,7 +8,16 @@
 
 import { i18next } from "@translations/invenio_communities/i18next";
 import React, { Component } from "react";
-import { Button, Icon, Modal, Message, Grid, Checkbox, Input } from "semantic-ui-react";
+import {
+  Button,
+  Icon,
+  Loader,
+  Modal,
+  Message,
+  Grid,
+  Checkbox,
+  Input,
+} from "semantic-ui-react";
 import PropTypes from "prop-types";
 import { Trans } from "react-i18next";
 import { communityErrorSerializer } from "../../api/serializers";
@@ -19,7 +28,7 @@ export class DeleteCommunityModal extends Component {
     super(props);
     this.INITIAL_STATE = {
       modalOpen: false,
-      loading: false,
+      loading: true,
       checkboxMembers: false,
       checkboxRecords: false,
       checkboxSlug: false,
@@ -36,8 +45,8 @@ export class DeleteCommunityModal extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { modalOpen } = this.state;
-    if (modalOpen && modalOpen !== prevState.modalOpen) {
+    const { loading, modalOpen } = this.state;
+    if (!loading && modalOpen && modalOpen !== prevState.modalOpen) {
       const {
         current: { inputRef },
       } = this.checkboxRef;
@@ -52,6 +61,7 @@ export class DeleteCommunityModal extends Component {
   }
 
   updateCommunityMetrics = async () => {
+    this.setState({ loading: true });
     const { community } = this.props;
     this.cancellableMembersCountFetch = withCancel(http.get(community.links.members));
     this.cancellableRecordsCountFetch = withCancel(http.get(community.links.records));
@@ -60,11 +70,16 @@ export class DeleteCommunityModal extends Component {
       const membersResponse = await this.cancellableMembersCountFetch.promise;
       const recordsResponse = await this.cancellableRecordsCountFetch.promise;
       this.setState({
+        loading: false,
         membersCount: membersResponse.data.hits.total,
         recordsCount: recordsResponse.data.hits.total,
       });
     } catch (error) {
       console.error(error);
+      const { message } = communityErrorSerializer(error);
+      if (message) {
+        this.setState({ error: message, loading: false });
+      }
     }
   };
 
@@ -98,7 +113,8 @@ export class DeleteCommunityModal extends Component {
   openConfirmModal = () => this.setState({ modalOpen: true });
 
   closeConfirmModal = () => {
-    this.setState(this.INITIAL_STATE);
+    let { loading } = this.state;
+    this.setState({ ...this.INITIAL_STATE, loading: loading });
     this.openModalBtnRef?.current?.focus();
   };
 
@@ -168,6 +184,7 @@ export class DeleteCommunityModal extends Component {
           size="tiny"
         >
           <Modal.Header>{i18next.t("Permanently delete community")}</Modal.Header>
+          {loading && <Loader active={loading} />}
           <Modal.Content>
             <Trans>
               Are you <strong>absolutely sure</strong> you want to delete the community?
@@ -190,8 +207,7 @@ export class DeleteCommunityModal extends Component {
                   id="members-confirm"
                   ref={this.checkboxRef}
                   label={
-                    /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
-                    <label>
+                    <label htmlFor="members-confirm">
                       <Trans>
                         <strong>{`${membersCount}`} members</strong> will be removed
                         from the community.
@@ -204,8 +220,7 @@ export class DeleteCommunityModal extends Component {
                 <Checkbox
                   id="records-confirm"
                   label={
-                    /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
-                    <label>
+                    <label htmlFor="records-confirm">
                       <Trans>
                         <strong>{`${recordsCount}`} records</strong> will be removed
                         from the community.
@@ -218,8 +233,7 @@ export class DeleteCommunityModal extends Component {
                 <Checkbox
                   id="slug-confirm"
                   label={
-                    /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
-                    <label>
+                    <label htmlFor="slug-confirm">
                       <Trans>
                         You <strong>CANNOT</strong> reuse the community identifier "
                         {{ communitySlug }}".
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
@@ -12,7 +12,7 @@ import { Button, Icon, Modal, Message, Grid, Checkbox, Input } from "semantic-ui
 import PropTypes from "prop-types";
 import { Trans } from "react-i18next";
 import { communityErrorSerializer } from "../../api/serializers";
-import { ErrorMessage, withCancel } from "react-invenio-forms";
+import { ErrorMessage, http, withCancel } from "react-invenio-forms";
 
 export class DeleteCommunityModal extends Component {
   constructor(props) {
@@ -31,6 +31,10 @@ export class DeleteCommunityModal extends Component {
     this.state = this.INITIAL_STATE;
   }
 
+  componentDidMount() {
+    this.updateCommunityMetrics();
+  }
+
   componentDidUpdate(prevProps, prevState) {
     const { modalOpen } = this.state;
     if (modalOpen && modalOpen !== prevState.modalOpen) {
@@ -43,7 +47,26 @@ export class DeleteCommunityModal extends Component {
 
   componentWillUnmount() {
     this.cancellableDelete && this.cancellableDelete.cancel();
+    this.cancellableMembersCountFetch && this.cancellableMembersCountFetch.cancel();
+    this.cancellableRecordsCountFetch && this.cancellableRecordsCountFetch.cancel();
   }
+
+  updateCommunityMetrics = async () => {
+    const { community } = this.props;
+    this.cancellableMembersCountFetch = withCancel(http.get(community.links.members));
+    this.cancellableRecordsCountFetch = withCancel(http.get(community.links.records));
+
+    try {
+      const membersResponse = await this.cancellableMembersCountFetch.promise;
+      const recordsResponse = await this.cancellableRecordsCountFetch.promise;
+      this.setState({
+        membersCount: membersResponse.data.hits.total,
+        recordsCount: recordsResponse.data.hits.total,
+      });
+    } catch (error) {
+      console.error(error);
+    }
+  };
 
   handleInputChange = (event) => {
     this.setState({ inputSlug: event.target.value });
@@ -111,6 +134,8 @@ export class DeleteCommunityModal extends Component {
       checkboxRecords,
       checkboxSlug,
       inputSlug,
+      membersCount,
+      recordsCount,
     } = this.state;
     const { label, community } = this.props;
     const communitySlug = community.slug;
@@ -168,8 +193,8 @@ export class DeleteCommunityModal extends Component {
                     /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
                     <label>
                       <Trans>
-                        <strong>All the members</strong> will be removed from the
-                        community.
+                        <strong>{`${membersCount}`} members</strong> will be removed
+                        from the community.
                       </Trans>
                     </label>
                   }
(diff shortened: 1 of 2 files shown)


invenio_communities/views/communities.py
```diff
@@ -187,6 +188,8 @@ def communities_settings(pid_value, community, community_ui):
     if type(max_size) is int and max_size > 0:
         logo_size_limit = max_size
 
+    members_count = MemberModel.count_members(community_id=community.id)
```
alejandromumo, line 191: This is querying directly the data layer in the presentation layer, it doesn't feel right.  

I suggest for this information to be hooked somewhere else (either service layer or maybe in the UISerializer?). The community is read from the service and then serialized [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/views/decorators.py#L28)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -13,6 +13,7 @@ import PropTypes from "prop-types";
 import { Trans } from "react-i18next";
 import { communityErrorSerializer } from "../../api/serializers";
 import { ErrorMessage, withCancel } from "react-invenio-forms";
+import axios from "axios";
```
alejandromumo, line 16: Note: It is not very common that we use axios directly like this, it's usually behind some kind of wrapper like in [React-Invenio-Deposit](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositApiClient.js#L30) or inside a search application (e.g. in communities search). 

Question: is the record count hookable to the community in the backend? If so, we could avoid making one request to the API just for this. Otherwise, ignore this comment :)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -168,8 +180,10 @@ export class DeleteCommunityModal extends Component {
                     /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
                     <label>
                       <Trans>
-                        <strong>All the members</strong> will be removed from the
-                        community.
+                        <strong>
+                          {membersCount ? membersCount.toString() : "All the"} members
```
alejandromumo, line 184: I am not sure how `<Trans>` deals with dynamic strings. E.g. this string could have any number or `All the` plus another string (that is always translated).

Since I am not an expert on this part, I am just leaving a note and will just let other jump in.

Also, I think the string `All the` can be removed since we're always passing the number of members to be deleted.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -25,12 +26,22 @@ export class DeleteCommunityModal extends Component {
...
+      recordsCount: false,
     };
     this.checkboxRef = React.createRef();
     this.openModalBtnRef = React.createRef();
     this.state = this.INITIAL_STATE;
   }
 
+  componentDidMount() {
+    const { community } = this.props;
+    axios.get(`/api/communities/${community.id}/records`).then((response) => {
```
kpsherva, line 38: the endpoint should not be hardcoded, we take it from `community.links`

kpsherva, line 38: the aync fetch should be wrapped in `withCancel` function, which helps us to avoid memory leaks in case of component being unmounted before the request finishes

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
kpsherva: ping @anikachurilova

invenio_communities/views/communities.py
kpsherva: I agree, this is not following the architecture of application layers, see https://inveniordm.docs.cern.ch/develop/architecture/software/#service-layer

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
anikachurilova: Looks valid for me. In any case, if not sure, you can always check how the string will be marked: https://inveniordm.docs.cern.ch/develop/howtos/i18n/#usage_1

kpsherva (commented):
LGTM, nice and cleaned!
I left two comments, but no action until Monday, until we discuss

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -25,10 +25,14 @@ export class DeleteCommunityModal extends Component {
...
       inputSlug: "",
       error: undefined,
+      membersCount: false,
+      recordsCount: false,
     };
     this.checkboxRef = React.createRef();
     this.openModalBtnRef = React.createRef();
     this.state = this.INITIAL_STATE;
+
+    this.calculateStats();
```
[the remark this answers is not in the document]
mitsosf: Note for Monday:
Was getting an ESLint warning when setting state in `componentDidMount` and that's why I moved it to the constructor.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -168,8 +193,8 @@ export class DeleteCommunityModal extends Component {
                     /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
                     <label>
                       <Trans>
-                        <strong>All the members</strong> will be removed from the
-                        community.
+                        <strong>{`${membersCount.toString()} members`}</strong> will be
```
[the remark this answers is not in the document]
mitsosf: Changed it to 

```javascript
<strong>{`${membersCount}`} members< /strong> will be
```

Otherwise, the `membersCount` value wouldn't print even if it was in the state. Another alternative is the `toString()` method.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -43,8 +47,27 @@ export class DeleteCommunityModal extends Component {
...
+    this.cancellableRecordsCountFetch && this.cancellableRecordsCountFetch.cancel();
   }
 
+  updateCommunityMetrics = async () => {
+    const { community } = this.props;
+    this.cancellableMembersCountFetch = withCancel(http.get(community.links.members));
+    this.cancellableRecordsCountFetch = withCancel(http.get(community.links.records));
+
+    try {
+      const membersResponse = await this.cancellableMembersCountFetch.promise;
```
kpsherva, line 60: from what I understand here, you are not handling the loading state while waiting on the response, what happens between opening the modal and receiving the reponse?

mitsosf: Should I set loading to `true` and not allow the user to click delete while they're being fetched?

I can also do something like 
```js
{membersCount && <Message.Content />}
```
essentially not showing the Message or Checkbox at all while the data loads.

kpsherva: > Should I set loading to `true` and not allow the user to click delete while they're being fetched?

yes, you should set loading to true (check other places on how it is done)

> I can also do something like
> 
> ```js
> {membersCount && <Message.Content />}
> ```
> 
> essentially not showing the Message or Checkbox at all while the data loads.

I think it would be better to show a nice loader while the user is waiting for response (instead of not showing anything). Check other places to see how to implement the loader.

mitsosf: After quite some effort to get the margins and paddings to work, I came up with this:
image

Let me know if you think it makes sense to leave it like this. The user will rarely have to see this render, if at any time.
I'm testing this by inserting this at L61:
`await new Promise(resolve => setTimeout(resolve, 5000));`
essentially before sending the requests.

kpsherva: for other reviewers: it can be tested also by setting a lower bandwidth in your browser's dev tools (f.e. to 3G)

kpsherva: > After quite some effort to get the margins and paddings to work, I came up with this: image
> 
> Let me know if you think it makes sense to leave it like this. The user will rarely have to see this render, if at any time. I'm testing this by inserting this at ~L61~ L68: `await new Promise(resolve => setTimeout(resolve, 5000));` essentially before sending the requests.

I think it is good, as long as you set a default height of the modal to match the final rendered one, in order to prevent the UI from resizing elements once the content is loaded (it jumps a bit). It is not a blocker though, can be done in a separate PR if you prefer

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -143,76 +183,87 @@ export class DeleteCommunityModal extends Component {
...
+                        </label>
+                      }
+                      checked={checkboxRecords}
+                      onChange={this.handleCheckboxChange}
+                    />
+                    <Checkbox
+                      id="slug-confirm"
+                      label={
+                        /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
+                        <label>
```
kpsherva, line 246:
```suggestion
                      label={
                        <label forHtml="slug-confirm">
```

kpsherva (approved):
LGTM! only two minor comments

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -43,8 +56,29 @@ export class DeleteCommunityModal extends Component {
...
+    try {
+      const membersResponse = await this.cancellableMembersCountFetch.promise;
+      const recordsResponse = await this.cancellableRecordsCountFetch.promise;
+      this.setState({
+        loading: false,
+        membersCount: membersResponse.data.hits.total,
+        recordsCount: recordsResponse.data.hits.total,
+      });
+    } catch (error) {
+      console.error(error);
```
kpsherva, line 78: missing loading: false on error, and displaying the error to the user (in the modal)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -75,6 +109,10 @@ export class DeleteCommunityModal extends Component {
   openConfirmModal = () => this.setState({ modalOpen: true });
 
   closeConfirmModal = () => {
+    let initialState = this.INITIAL_STATE;
+    let { loading } = this.state;
+    initialState.loading = loading;
```
kpsherva, line 114: I am not fully understanding this state manipulation here. Why do we need to modify initial state ?

mitsosf: Because I want to maintain the current loading status, because the initial state `loading` is `false`. 
The component doesn't get unmounted when we hit `Close` or lose focus of it.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
mitsosf: Will pring 'All' in case of network/request/undefined error, because changing the content looks complicated.

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -43,8 +56,29 @@ export class DeleteCommunityModal extends Component {
...
   }
 
+  updateCommunityMetrics = async () => {
+    this.setState({ loading: true });
+    const { community } = this.props;
+    this.cancellableMembersCountFetch = withCancel(http.get(community.links.members));
+    this.cancellableRecordsCountFetch = withCancel(http.get(community.links.records));
+
+    try {
+      const membersResponse = await this.cancellableMembersCountFetch.promise;
```
ntarocco, line 70: Suggestion to make them in parallel:
```js
const [membersResponse, recordsResponse] = await Promise.all([call1(), call2()]);
```
to be tested :)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -8,18 +8,27 @@
...
-import { ErrorMessage, withCancel } from "react-invenio-forms";
+import { ErrorMessage, http, withCancel } from "react-invenio-forms";
 
 export class DeleteCommunityModal extends Component {
   constructor(props) {
     super(props);
     this.INITIAL_STATE = {
       modalOpen: false,
-      loading: false,
+      loading: true,
```
ntarocco, line 31:
```suggestion
      loading: true,
      membersCount: undefined,
      recordsCount: undefined
```

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
ntarocco, line 78: I don't think it is enough... it will not display anything to the user, no?
We should set the `error` with the message in the state (and test to see how it looks).

ntarocco (commented):
Looks very nice, added a couple of comments :)

invenio_communities/assets/semantic-ui/js/invenio_communities/settings/profile/DeleteCommunityModal.js
```diff
@@ -43,8 +56,29 @@ export class DeleteCommunityModal extends Component {
...
     this.cancellableDelete && this.cancellableDelete.cancel();
+    this.cancellableMembersCountFetch && this.cancellableMembersCountFetch.cancel();
+    this.cancellableRecordsCountFetch && this.cancellableRecordsCountFetch.cancel();
   }
 
+  updateCommunityMetrics = async () => {
+    this.setState({ loading: true });
+    const { community } = this.props;
+    this.cancellableMembersCountFetch = withCancel(http.get(community.links.members));
+    this.cancellableRecordsCountFetch = withCancel(http.get(community.links.records));
```
rekt-hard, line 67: Since we are only interested in the total, would it make sense to limit the number of results? Passing `?s=1` or even `s=0`should be enough, so the backend does not have to serialize 10 records (as per default).
This request will probably not be performed very often but just in case.

PR #945, opened 2023-04-19
cache: adds unmanaged groups to be cached and loaded in the identity

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2186


jrcastro2:
tests passing locally

invenio_communities/utils.py
```diff
@@ -64,6 +70,15 @@ def on_membership_change(identity=None):
         current_cache.delete(identity_cache_key(identity))
 
 
+def delete_all_user_communities_keys():
+    """Handler called when a group is updated/added/removed from a community."""
+    binary_keys = current_cache.cache._read_client.keys(f"*{IDENTITY_KEY}*")
```
jrcastro2, line 75: This is not ideal, but there is no way to fetch concrete keys from flask-caching. That's why we are accessing protected methods.

invenio_communities/cache/cache.py
```diff
@@ -0,0 +1,44 @@
...
+    @cached_property
+    def timeout(self):
+        """Return default timeout from config."""
+        return current_app.config["COMMUNITIES_IDENTITIES_CACHE_TIME"]
+
+    def get(self, key):
+        """Return the key value.
+
+        :param key: the object's key
+        """
```
ntarocco, line 29: maybe a good idea to add `pass` or `raise NotImplementedError`? or maybe mark them as `abstract`?

invenio_communities/cache/cache.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Abstract simple identity cache definition."""
+from flask import current_app
+from werkzeug.utils import cached_property
+
+
+class IdentityCache(object):
```
ntarocco, line 14:
```suggestion
class IdentityCache:
```

invenio_communities/cache/redis.py
```diff
@@ -0,0 +1,54 @@
...
+
+from invenio_communities.cache.cache import IdentityCache
+
+
+class IdentityRedisCache(IdentityCache):
+    """Redis image cache."""
+
+    def __init__(self, app=None):
+        """Initialize the cache."""
+        super(IdentityRedisCache, self).__init__(app=app)
```
ntarocco, line 22:
```suggestion
        super().__init__(app=app)
```

invenio_communities/cache/redis.py
```diff
@@ -0,0 +1,54 @@
...
+        app = app or current_app
+        redis_url = app.config["COMMUNITIES_IDENTITIES_CACHE_REDIS_URL"]
+        prefix = app.config.get("COMMUNITIES_IDENTITIES_CACHE_REDIS_PREFIX", "identity")
+        self.cache = RedisCache(host=StrictRedis.from_url(redis_url), key_prefix=prefix)
```
ntarocco, line 26: Is it, in the end, necessary to bypass the cache abstraction and go directly to Redis? Is it because we discussed separating in different cache databases, so that when it is cleaned, only that db is touched?

invenio_communities/ext.py
```diff
@@ -10,7 +10,10 @@
 """Invenio communities extension."""
 
 from flask_principal import identity_loaded
+from invenio_accounts.signals import datastore_post_commit
 from invenio_records_resources.services import FileService
+from six import string_types
```
ntarocco, line 15: `six` is not needed any more :) Please remove.

invenio_communities/members/services/components.py
```diff
@@ -50,20 +47,31 @@ def _get_user_ids_from_group_id(role_id):
...
 
     def accept_invite(self, identity, record=None, data=None, **kwargs):
         """On accept invite."""
         self._member_changed(record)
 
-    def members_add(self, identity, record=None, data=None, **kwargs):
+    def members_add(self, identity, record=None, community=None, data=None, **kwargs):
         """On member add (only for groups)."""
-        self._member_changed(record)
+        if record.get("type") == "group":
```
ntarocco, line 58: is `type` mandatory? Or optional?

invenio_communities/members/services/components.py
```diff
@@ -50,20 +47,31 @@ def _get_user_ids_from_group_id(role_id):
...
+            role = Role.query.filter_by(id=record["id"]).one_or_none()
+            if role.is_managed:
+                users = role.users.all()
+                for user in users:
+                    on_user_membership_change(Identity(user.id))
```
ntarocco, line 63: Here there is room for improvement:
- we get the user id
- we wrap it around an Identity
- then, when deleting, we get again the `id`

invenio_communities/members/services/schemas.py
```diff
@@ -140,7 +140,8 @@ def get_group_member(self, group):
         return {
             "type": "group",
             "id": group["id"],
-            "name": group["id"] or "",
+            "name": group.get("name") or group["id"],
```
ntarocco, line 143: is the name optional or mandatory in the DB?

invenio_communities/utils.py
```diff
@@ -8,10 +8,31 @@
...
+
+def add_to_community_cache(community_id, identity_id):
+    """Updates the community cache."""
+    community_cache = current_identities_cache.get(community_id)
+    if community_cache and identity_id not in community_cache:
+        community_cache.append(identity_id)
+    else:
+        community_cache = [identity_id]
+    # This cache is set to the same duration as the identity caches, because it should not expire before
+    # This cache has a risk of be eternal therefore a cronjob that clears all the cache has to be set on a daily basis
```
ntarocco, line 30:
```suggestion
    # This cache has a risk of being eternal therefore a cronjob that clears all the cache has to be set on a daily basis
```

invenio_communities/members/services/components.py
jrcastro2: It`s mandatory, will fix it, thanks!

invenio_communities/members/services/schemas.py
jrcastro2: Name is optional [see here](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L69)

invenio_communities/alembic/37b21951084c_update_role_id_type.py
```diff
@@ -0,0 +1,95 @@
...
+from alembic import op
+from invenio_accounts.alembic.f2522cdd5fcd_change_accountsrole_primary_key_to_ import (
+    downgrade as accounts_downgrade,
+)
+
+# revision identifiers, used by Alembic.
+revision = "37b21951084c"
+down_revision = "a14fa442680f"
+branch_labels = ()
+depends_on = "f2522cdd5fcd"
```
alejandromumo, line 19: question: where is this revision coming from? I can't find it in invenio org.

invenio_communities/alembic/37b21951084c_update_role_id_type.py
```diff
@@ -0,0 +1,95 @@
...
+"""Update role_id type."""
+import sqlalchemy as sa
+from alembic import op
+from invenio_accounts.alembic.f2522cdd5fcd_change_accountsrole_primary_key_to_ import (
+    downgrade as accounts_downgrade,
+)
+
+# revision identifiers, used by Alembic.
+revision = "37b21951084c"
+down_revision = "a14fa442680f"
```
alejandromumo, line 17: question: this revision is from [invenio-requests](https://github.com/inveniosoftware/invenio-requests/blob/18594a369a0c568b2c83a14ed9e1cb3fef447ae8/invenio_requests/alembic/a14fa442680f_create_tables.py#L15). Shouldn't it be `depends_on=a14fa442680f`?

invenio_communities/cache/redis.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def set(self, key, value, timeout=None):
+        """Cache the object.
+
+        :param key: the object's key
+        :param value: the stored object
+        :type value: `BytesIO` object
+        :param timeout: the cache timeout in seconds
+        """
+        timeout = timeout or self.timeout
```
alejandromumo, line 45: note: some keys might need to be deleted maybe? Passing `None` would be enough?

invenio_communities/utils.py
```diff
@@ -8,10 +8,31 @@
...
+from invenio_accounts.proxies import current_db_change_history
 
 from .generators import CommunityRoleNeed
-from .proxies import current_communities
+from .proxies import current_communities, current_identities_cache
+
+IDENTITY_KEY = "user-communities:"
+
+
+def add_to_community_cache(community_id, identity_id):
```
alejandromumo, line 22: shelve: we're adding some data fetching logic (from the cache) in `utils`, does not feel right. However it can be tackled afterwards.

invenio_communities/members/services/components.py
```diff
@@ -14,32 +14,29 @@
...
-            r = Role.query.filter(Role.id == role_id).one()
-            return [u.id for u in r.users.all()]
-
         user_ids = []
         if isinstance(member, MemberMixin):
             if member.user_id:
                 user_ids = [member.user_id]
             elif member.group_id:
-                user_ids = _get_user_ids_from_group_id(member.group_id)
+                on_group_membership_change(str(community.id))
```
alejandromumo, line 32: major: `community` must be defined. Otherwise it can raise an exception. 


question: `member.group_id` wiped before calling this component?

tests/cache/test_identity_redis_cache.py
```diff
@@ -0,0 +1,39 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Image Redis Cache Tests."""
+
+from __future__ import absolute_import
+
+from invenio_communities.proxies import current_identities_cache
+
+
+def test_cache_deletion(app):
```
alejandromumo, line 16: Idea: we can take inspiration for more tests from [here](https://github.com/inveniosoftware/invenio-rdm-migrator/blob/master/tests/streams/test_cache.py) 😇

tests/conftest.py
```diff
@@ -58,6 +58,15 @@ def app_config(app_config):
         "R": "Remote",
     }
     app_config["FILES_REST_DEFAULT_STORAGE_CLASS"] = "L"
+    app_config["COMMUNITIES_IDENTITIES_CACHE_TIME"] = 60 * 60 * 24
```
alejandromumo, line 61: Maybe we can test the expires (e.g. small timeouts)

invenio_communities/members/services/service.py
```diff
@@ -241,15 +242,6 @@ def _add_factory(
         request_id=None,
     ):
         """Add a member to the community."""
-        # TODO: inefficient to do here, should be done in bulk instead.
-        if member["type"] == "group":
```
alejandromumo, line 245: question: where did this code go?

alejandromumo (changes requested):
left multiple questions

ntarocco (commented):
Can be merged after discussed changes and alembic.

invenio_communities/alembic/37b21951084c_update_role_id_type.py
jrcastro2: Fixed, was a leftover

invenio_communities/config.py
```diff
@@ -293,3 +293,15 @@
 
 # Preview: Feature flag for invenio-app-rdm v11
 COMMUNITIES_ADMINISTRATION_DISABLED = True
+
+# Cache duration
+# 60 seconds * 60 (1 hour) * 24 (1 day)
+COMMUNITIES_IDENTITIES_CACHE_TIME = 60 * 60 * 24
+
+# Redis URL Cache for identities
+COMMUNITIES_IDENTITIES_CACHE_REDIS_URL = "redis://localhost:6379/3"
```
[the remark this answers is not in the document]
jrcastro2: 3 is used because 0, 1 and 2 are already being used. Since this is a config variable it is configurable already.

invenio_communities/cache/redis.py
jrcastro2: By passing `None` there is no timeout and the keys are not deleted. you can set the value in seconds and after the set time they will be deleted. Otherwise, if we need to delete a key we can use the delete func below.

invenio_communities/members/services/service.py
jrcastro2: This was a hack to map the name to the id. Not any more needed. Now you can have a role with an ID which is different from the name.

invenio_communities/utils.py
```diff
@@ -43,27 +64,69 @@ def load_community_needs(identity):
...
-        identity.provides.add(CommunityRoleNeed(c_id, role))
+        for community_id, role in community_roles:
+            add_to_community_cache(community_id, identity.id)
+            # Add community needs to identity
+            identity.provides.add(CommunityRoleNeed(community_id, role))
+
+        current_identities_cache.set(
+            cache_key,
+            community_roles,
+            timeout=current_app.config["COMMUNITIES_IDENTITIES_CACHE_TIME"],
```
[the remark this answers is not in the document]
jrcastro2: Yes, we can remove it :+1:

invenio_communities/utils.py
jrcastro2: Not sure why `utils` os not right. More than fetching we are setting some values in the cache but every time we set we need to fetch them to add the new element to the list as the set func from the cache does not allow to "append". I guess we could add an `add` func in the RedisCache class that does this logic (but it would assume that the value of the key is a list/set).

invenio_communities/members/services/components.py
jrcastro2: This function takes care of updating the members, for both `user` and `groups`. However, for users we do not need to pass the community but for groups we need. If we try to upgrade a group wihout passing the community it has to raise an exception. Alternatively we can split the functions to differentiate the change of membership between `users` and `groups`, but if we try to update a group without passing a community it will always raise an exception as it has to.

jrcastro2: > question: member.group_id wiped before calling this component?

As far as I have tested it's not wiped before calling the component. Why do you think it's being wiped before?

tests/conftest.py
jrcastro2: Right! Good point! thanks :)

tests/cache/test_identity_redis_cache.py
jrcastro2: I have improved the tests with your suggestion of testing the default timeout expiration. I think all cases are now covered unless I am missing something

tests/members/test_members_services.py
```diff
@@ -127,9 +128,9 @@ def test_add_visible_property(member_service, community, owner, new_user, db):
 def test_add_invalid_data(member_service, community, owner, group, db):
     """Test various forms of invalid data."""
     # Invalid group id
-    data = {"members": [{"type": "group", "id": "invalid"}], "role": "reader"}
+    data = {"members": [{"type": "group", "id": 123}], "role": "reader"}
```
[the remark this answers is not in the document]
jrcastro2: We've updated the fixture that creates the tests to use the new type (String), as long as the uuids are converted to strings it's OK, I think it's enough if the tests check the type of the id, independently of how the id was generated (as long as it's a string), unless I am missing something

invenio_communities/alembic/fbe746957cfc_create_member_tables.py
```diff
@@ -14,9 +14,12 @@
 
 # revision identifiers, used by Alembic.
 revision = "fbe746957cfc"
-down_revision = "f701a32e6fbe"
+down_revision = (
+    "2f63be7b7572",
+    "f701a32e6fbe",
```
jrcastro2, line 19: MAJOR: Question for reviewers, with this change we are able to use the upgrade command from alembic even from scratch, and I think it's correct that this dependency directly depends on the recipe that creates. (that revision id could potentially be moved to `depends_on`, to be tested if )

slint: To me this sounds like the "alembic merge" is the way to enforce the order we were discussing IRL. We should do that at least for the recipes requiring this explicitness, i.e. PK type changes...

Now for the rest, my gut feeling is that we should also fix them in the long term, but it looks like that's a much bigger  task... Do we know roughly how many places that is? Is there a way to incrementally fix them across modules?

jrcastro2: To me the "alembic merge"  is the solution, only if we need to downgrade to that recipe, or to force the path through that recipe, the `depends_on` does not enforce that path but just tells  that it needs the revision_id to be processed before being executed (meaning that many others recipes could be executed in between both recipes).

About how many places should we fix it ... So far it looks like it works with this one, however there might be other places that are affected. To find them, we should review all the recipes, to fix them incrementally, we can just fix them as they start to fail :sweat_smile:

ppanero (commented):
copyright years headers 😇

invenio_communities/alembic/fbe746957cfc_create_member_tables.py
kpsherva: discussed in person, decided to go ahead with the solution and mitigate any consequences it might cause

invenio_communities/members/services/components.py
kpsherva: I agree with Manuel, you have set the community to be None as default, maybe we need to raise an exception to check if it is passed? or in the conditional statement we need to check if community is not None?

invenio_communities/utils.py
kpsherva: I agree, it does not feel right, because I wouldn't expect from utility function to be obliged to have application context (and current_identities_cache kinda indicates that), but I can see it was done like this previously...
In the end you are checking for duplicates in the cache, why not making the .set method of the cache class to check it ?

invenio_communities/utils.py
```diff
@@ -43,27 +63,64 @@ def load_community_needs(identity):
...
+    community_roles = current_identities_cache.get(cache_key)
     if community_roles is None:
         # aka Member.get_memberships(identity)
-        community_roles = (
-            current_communities.service.members.config.record_cls.get_memberships(
-                identity
-            )
+        group_ids = session.get("_unmanaged_groups", [])
+
+        member_cls = current_communities.service.members.config.record_cls
```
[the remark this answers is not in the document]
kpsherva: agree

invenio_communities/utils.py
```diff
@@ -43,27 +63,64 @@ def load_community_needs(identity):
...
+    """Clears the cache for the user identity."""
+    sid = id(session)
+    if current_db_change_history.sessions.get(sid):
+        for user_id in current_db_change_history.sessions[sid].updated_users:
+            on_user_membership_change(Identity(user_id))
+
+        for user_id in current_db_change_history.sessions[sid].deleted_users:
+            on_user_membership_change(Identity(user_id))
+
+        for role_id in current_db_change_history.sessions[sid].deleted_roles:
```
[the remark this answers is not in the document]
jrcastro2: Updating roles is only related to name or description changes of the role which will have no impact on it's members, nor permissions to access a community

invenio_communities/utils.py
```diff
@@ -43,27 +63,64 @@ def load_community_needs(identity):
...
+def on_user_membership_change(identity=None):
+    """Handler called when a user membership is changed."""
     if identity is not None:
-        current_cache.delete(identity_cache_key(identity))
+        current_identities_cache.delete(identity_cache_key(identity))
+
+
+def on_group_membership_change(community_id):
+    """Handler called when a group membership is changed."""
+    community_identities_cache = current_identities_cache.get(community_id)
```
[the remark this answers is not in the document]
jrcastro2: The cache get does not allow to return a default if missing

PR #954, opened 2023-05-05
notifications: add member recipient generator

rekt-hard:
tests: add notification member recipient generator test case

### Description

refs https://github.com/inveniosoftware/invenio-rdm-records/pull/990

Add member recipient generator for notifications.

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

2. You agree that you have the right to license your contribution under the current repository’s license.


invenio_communities/members/services/service.py
```diff
@@ -328,16 +328,27 @@ def _invite_factory(self, identity, community, role, visible, member, message, u
...
+        identity,
+        community_id,
+        params=None,
+        search_preference=None,
+        extra_filter=None,
+        **kwargs
     ):
         """Search."""
+        # Apply extra filters
+        filter_ = dsl.Q("term", **{"active": True})
```
kpsherva, line 341: I am not sure if it was meant here, see that there is a service method below which has similar filtering, I think this one was mean to search for all members, but maybe someone else can confirm as well

ntarocco: To be more clear, it looks like that this method and the new `scan` below are exactly the same?

invenio_communities/members/services/service.py
```diff
@@ -420,7 +459,7 @@ def _members_search(
             extra_filter=filter,
             **kwargs
         )
-        search_result = search.execute()
+        search_result = search.scan() if scan else search.execute()
```
kpsherva, line 462: to consider, not necessarily applying here (how many community members we might have?)
the scan might be problematic without setting timeout for scroll 
(see usage in https://github.com/CERNDocumentServer/cds-ils/blob/215c8c75790a6493b1365c7d47633e9df1cd4eae/cds_ils/migrator/relations/series.py#L63) 
the scan timeouts after a while, see https://opensearch.org/docs/latest/api-reference/scroll/

invenio_communities/notifications/generators.py
```diff
@@ -0,0 +1,54 @@
...
+
+        members = current_communities.service.members.scan(
+            system_identity,
+            community["id"],
+            extra_filter=filter_,
+        )
+
+        user_ids = []
+        for m in members:
+            if m["member"]["type"] != "user":
```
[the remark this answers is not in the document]
kpsherva: I think it could be improved in a separate task after we integrate groups? what do you think? ping @jrcastro2 @ntarocco

invenio_communities/members/services/service.py
rekt-hard: Thank you. The `scan=True` parameter is missing in the `scan`method. Will add in a bit.

It discussed with @slint and @kpsherva to be more explicit about the usage of scan. So not only passing `scan=True` to the `search()` method but explicitly calling `scan()`

invenio_communities/notifications/generators.py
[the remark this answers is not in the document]
rekt-hard: At the time of the RFC and implementation, I believe the groups were not fully integrated.
It should be supported in a future update though. Will add a TODO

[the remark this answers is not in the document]
ntarocco: After an internal discussion, sending notifications to groups is out of context. :)

invenio_communities/members/services/service.py
ntarocco: Indeed, it should have a sane default, overridable via args.

rekt-hard: There is a default timeout set to 5 minutes: https://github.com/opensearch-project/opensearch-py/blob/fe3b5a8922aa8eb04f735c74d127d7ea68a00bec/opensearchpy/helpers/actions.py#L492-L503

As I understand, the scroll timeout will also be reset after each "page" (all hits for the current scroll action have been exhausted): https://github.com/opensearch-project/opensearch-py/blob/fe3b5a8922aa8eb04f735c74d127d7ea68a00bec/opensearchpy/helpers/actions.py#L598-L600
As an example, calling scan will:
- perform a scroll search
- fetch 1000 hits (out of 2000)
- keep the state alive for 5 minutes
- yield hit for hit of current hits
- after exhausting the current hits, start again, until there are no more new hits available

It should also already be overridable via something like `search_preference={"scroll": "4h"}`. Though I have not tested this.

Do you think the default 5 minutes is too much? For OAI set percolators, we have set the default to 1 minute:
- https://github.com/inveniosoftware/invenio-oaiserver/blob/42ed11d1963ce2b0a6a18cec99b970e27c9a08ce/invenio_oaiserver/percolator.py#L152-L157
- https://github.com/inveniosoftware/invenio-oaiserver/blob/42ed11d1963ce2b0a6a18cec99b970e27c9a08ce/invenio_oaiserver/config.py#L55

invenio_communities/members/services/service.py
kpsherva: lets discuss this during standup, I am not sure I understand why we need the scan=True parameter in the scan method.

invenio_communities/members/services/service.py
kpsherva: discussed IRL, agreed on adding a comment reminding about the default scroll time

invenio_communities/members/services/service.py
```diff
@@ -420,14 +460,16 @@ def _members_search(
...
 
         return self.result_list(
             self,
             identity,
             search_result,
             params,
-            links_tpl=LinksTemplate(
+            links_tpl=None
+            if scan
+            else LinksTemplate(
```
kpsherva, line 472: why do we need to return no links if we scan ?

rekt-hard: Links for each ResultItem will be created.
Links for the ResultList are not possible, as the pagination depends on the total hits of the query. When using scan, the hits object is only a generator and no information about total number of results is available (https://github.com/inveniosoftware/invenio-records-resources/blob/5335294dade21decea0f527022d96e12e1ffad52/invenio_records_resources/services/records/results.py#L165-L172)

The tests should cover dumping the result of scan and search, and checking if all dumped items are the same (including links)

invenio_communities/members/services/service.py
rekt-hard: did a bit of digging and testing. seems like the scroll params are not overridable so easily (sorry about that info prior).

We could add explicit `scan_params` to the `_members_search` method, and depending on the scan flag, add these params to the search before starting the scan.

```py
    def _members_search(...,
        scan=False,
        # add explicit scan_params
        scan_params=None,
        **kwargs
    ):
        """Members search."""
        ...
        # Prepare and execute the search
        params = params or {}
        # set explicit scan_params
        scan_params = scan_params or {}

        search = self._search(
            "search_members",
            identity,
            params,
            search_preference,
            search_opts=search_opts,
            permission_action=None,
            extra_filter=filter,
            **kwargs
        )

        # use explicit scan_params
        search_result = search.params(**scan_params).scan() if scan else search.execute()
```

This way, we can override the params passed to scroll. What do you think?

invenio_communities/notifications/generators.py
```diff
@@ -0,0 +1,55 @@
...
+            # TODO: add support for groups
+            if m["member"]["type"] != "user":
+                continue
+            user_ids.append(m["member"]["id"])
+
+        if not user_ids:
+            return recipients
+
+        filter_ = dsl.Q("terms", **{"id": user_ids})
+        users = current_users_service.scan(system_identity, extra_filter=filter_)
```
slint, line 52: minor/question: I assume we do this extra query because the serialization from community members doesn't include the full information/schema of a user?

slint (approved):
One question about a second search call to users, but otherwise LGTM.

PR #959, opened 2023-05-16
assets: configure number of items in carousel

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/54

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
@@ -47,12 +47,14 @@ class CommunitiesCarousel extends Component {
   }
 
   getDataIndex = (index) => {
+    const { itemsPerPage } = this.props;
     const {
       data: { hits },
     } = this.state;
-    if (index > hits.length - 1) return 0;
-    if (index < 0) return hits.length - 1;
-    return index;
+    const i = index * parseInt(itemsPerPage);
+    if (i > hits.length - 1) return 0;
+    if (i < 0) return hits.length - 1;
+    return i;
   };
 
   runCarousel = async (newIndex) => {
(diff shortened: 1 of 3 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
```diff
@@ -38,12 +39,14 @@ class CommunitiesCarousel extends Component {
...
     const {
       data: { hits },
     } = this.state;
-    if (index > hits.length - 1) return 0;
-    if (index < 0) return hits.length - 1;
-    return index;
+    const i = index * parseInt(itemsPerPage);
+    if (i > hits.length - 1) return 0;
+    if (i < 0) return hits.length - 1;
+    return i;
```
jennur, line 49: I think this could potentially leave some items out? E.g. if `hits.length = 5` and `itemsPerPage = 2`, then `index = 3` will return 0 and the 5th item will never show. Maybe we could do something like this instead (not tested): 

```javascript
    const pages = Math.ceil(hits.length / parseInt(itemsPerPage));
    if (index > pages) return 0;
    if (index < 0) pages - 1;
    return index;
```

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
```diff
@@ -80,28 +83,41 @@ class CommunitiesCarousel extends Component {
...
       console.error(error);
     }
   };
 
+  carouselSlides = () => {
+    const { data, activeIndex } = this.state;
+    const { defaultLogo, itemsPerPage } = this.props;
+
+    const sliceEnd = parseInt(activeIndex) + parseInt(itemsPerPage);
+    console.log(sliceEnd, "================ end");
```
jennur, line 101: Test leftovers?

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
jrcastro2: I think that's correct. If `hits.length = 5` and you want to display the fifth element then the `index` value should be 2 and not 3, if index is 3 it would display elements 7 and 8. If I understand correctly, `itemsPerPage` indicates the amount of items displayed and for each value of the `index`, we display as many elements as indicated in the `itemsPerPage`.

jrcastro2 (commented):
LGTM!
Only eslint is complaining and some leftovers as pointed by Jenny.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CommunitiesCarousel.js
jennur: You're right! My bad, I wasn't considering the 0 index 😄

PR #961, opened 2023-05-23
CarouselItem: escape html

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2216

## Screenshot
Screenshot 2023-05-23 at 13 42 09

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CarouselItem.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/CarouselItem.js
@@ -54,7 +54,10 @@ class CarouselItem extends Component {
               </Grid.Column>
             </Item.Header>
             <Item.Description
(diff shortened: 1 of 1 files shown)


lnielsen:
This is a great example of why having more reviews on the code is important! We would have created a Cross-Site Scripting (XSS) vulnerability in the code if ``SanitizedUnicode`` wasn't changed to ``SanitizedHTML``. There's absolutely no blame on anyone in my statement, rather I see it as a success that our reviews are actually working and we caught the issue before it made it to a production system.

jennur:
> I think that you will also have to change the sanitization of the `description` field: it currently uses `SanitizedUnicode`, we should probably change it to `SanitizedHTML`: https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/schema.py#L114 Can you check and maybe check also the tests?

Thanks a lot! I wasn't aware of this. Will fix!

Issue #874, opened 2023-01-26
refactor: featured communities macros


kpsherva:
The recent uploads and community carousel macros need to be rewritten to avoid using react (since it is not needed). Due to the long loading time the react is not very attractive to be placed on the frontpage.
On the other hand it could be nice not to delete the react code entirely - since it can be potentially used on the other pages.

Steps to complete (divided by PRs to create):

1. Add CSS loading placeholders to community carousel and recent uploads react components (to make the demo sites more attractive when loading) (core invenioRDM). How to implement them check for the community management side panel box on the record landing page (before react loads, the CSS placeholder is shown, then the react placeholder takes over until the information is fetched).
2. Apply the same changes on the zenodo-rdm and cds-rdm (and who-rdm if needed - if anything is overridden there)

3. Rewrite the components in pure HTML and CSS. It will need adding jinja context variables to the frontpage view, which is located in invenio-theme.

from Nico:
invenio-theme by default does not register the endpoint, see [here](https://github.com/inveniosoftware/invenio-theme/blob/maint-1.1/invenio_theme/ext.py#L60) and related config [here](https://github.com/inveniosoftware/invenio-theme/blob/maint-1.1/invenio_theme/config.py#L97)

Zenodo they have heavily customized it, adding the cache too, see the entire module [here](https://github.com/zenodo/zenodo/tree/master/zenodo/modules/frontpage).
in CDS Videos, we have done something very similar, see [here](https://github.com/CERNDocumentServer/cds-videos/tree/main/cds/modules/home).

In order to add the custom frontpage view, you can follow the previous implementation.

PR #966, opened 2023-06-02
communities carousel: rewrite React to jinja & jquery

jennur:
Closes https://github.com/inveniosoftware/invenio-communities/issues/874

### Known issues
- The left animation is not working correctly when switching between slides, except 2 -> 0
- The right animation is buggy between index 0 -> 2
- The carousel item needs a placeholder for non-existing logos

**SUI transition docs** https://semantic-ui.com/modules/transition.html#/definition


invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
slint, line 1: minor: JS-noob question: doesn't all this need to be wrapped in a `$(document).ready(function () { ... });`? Or are we fine since the `<script>` tag is at the end of the page?

invenio_communities/templates/semantic-ui/invenio_communities/macros/carousel_item.html
slint, line 1: minor/nit: do we have a convention for splitting each macro in its own separate .html file? E.g. would this be useful for someone that wants to override the macro/template?

slint (approved):
LGTM! Minor comments on file placement and if we need any special JS to start the carousel.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
```diff
@@ -5,31 +5,67 @@
  * Invenio is free software; you can redistribute it and/or modify it
  * under the terms of the MIT License; see LICENSE file for more details.
  */
-import React from "react";
-import ReactDOM from "react-dom";
-import CommunitiesCarousel from "./CommunitiesCarousel";
-import { OverridableContext, overrideStore } from "react-overridable";
+import $ from "jquery";
```
ntarocco, line 8: annoying guy mode on :)
Can we get rid of all the JS and use CSS only? [Example](https://css-tricks.com/css-only-carousel/) first link searching on the internet...

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
ntarocco: Indeed, the script tag will not be added at the end of the page, but inside the macro, which could be added anywhere. If we can remove JS completely, even better :)

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
slint: Agree on that if it's possible. 

An alternative, in case it's too difficult/tricky to get right, would be to optimize for how the page loading looks like with "Slow 3G" throttling. If we get the first carousel "slide" to display ok, and get the interactivity later, maybe it's good enough? That's how it currently works for the legacy Zenodo frontpage.

ntarocco: Tip: to achieve that, we can try to avoid any dependency and simply use plain JS (zero imports, no jquery). In this way, we will avoid depending on any bundle, and the JS can be downloaded and executed independently of other bundles.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
```diff
@@ -5,31 +5,67 @@
...
-      intervalDelay={intervalDelay}
-      animationSpeed={animationSpeed}
-      defaultLogo={defaultLogo}
-      itemsPerPage={itemsPerPage}
-    />
-  </OverridableContext.Provider>,
-  communitiesCarouselContainer
-);
+
+const carouselSlides = $(".carousel.transition");
```
ntarocco, line 14: would it be possible to use `#` instead of `.`? Otherwise, if `.carousel.transition` is a child of an id, `$("#id").find(".carousel.transition")` is much more performant. (I used jQuery here, but we should do it without.

invenio_communities/templates/semantic-ui/invenio_communities/macros/carousel_item.html
jennur: We don't have a convention as far as I know, I guess it's more personal preference. I find it easier to navigate this way, but if there are arguments against it, let me know.

invenio_communities/assets/semantic-ui/js/invenio_communities/community/communitiesCarousel/index.js
jennur: Testing with (fast) 3g (slow 3g takes forever to show anything at all - just a white page) the template is loading equally fast as the temporary "recent uploads" placeholders, but without any functionality for a while, so I guess that's what you suggested already?

invenio_communities/templates/semantic-ui/invenio_communities/macros/carousel_item.html
```diff
@@ -0,0 +1,43 @@
+{# -*- coding: utf-8 -*-
+
+  This file is part of Invenio.
+  Copyright (C) 2023 CERN.
+
+  Invenio is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% macro carousel_item(community=None) %}
```
zzacharo, line 10: I would keep here just the initial macro and add any new code to any instance that implements this e.g zenodo-rdm

jennur:
Moved to `zenodo-rdm` https://github.com/zenodo/zenodo-rdm/pull/398

PR #996, opened 2023-08-14
details header: remove hardcoded header, remove empty tabs

jennur:
Closes https://github.com/zenodo/rdm-project/issues/172, closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2327

- Removes "about" and/or "curation policy" tabs if no content
- Removes hard coded headline for Curation policy

No page contents added:
Screenshot 2023-08-14 at 15 34 35
Only curation policy content added:
Screenshot 2023-08-14 at 15 35 11
Both added:
Screenshot 2023-08-14 at 15 35 41

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
@@ -139,7 +139,7 @@
     <div class="ui divider mobile only"></div>
     <div
       class="ui container secondary pointing stackable menu page-subheader pl-0 pr-0">
-      {% for item in current_menu.submenu('communities').children if item.permissions == True or permissions[item.permissions] %}
+      {% for item in current_menu.submenu('communities').children if (item.permissions == True or permissions[item.permissions]) and item.visible %}
           <a
             class="item {{ 'active' if active_community_header_menu_item == item.name }} {{ 'disabled' if not item.url }}"
             href="{{ item.url }}"
--- a/invenio_communities/views/decorators.py
+++ b/invenio_communities/views/decorators.py
@@ -10,7 +10,7 @@
 
 from functools import wraps
 
-from flask import g
+from flask import g, request
 
 from invenio_communities.communities.resources.serializer import (
     UICommunityJSONSerializer,
@@ -29,6 +29,7 @@ def pass_community(serialize):
                 id_=pid_value, identity=g.identity
             )
             kwargs["community"] = community
+            request.community = community.to_dict()
             if serialize:
                 community_ui = UICommunityJSONSerializer().dump_obj(community.to_dict())
                 kwargs["community_ui"] = community_ui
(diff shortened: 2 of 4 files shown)


invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -140,13 +140,19 @@ <h1 class="ui medium header mb-0">{{ community.metadata.title }}</h1>
     <div
       class="ui container secondary pointing stackable menu page-subheader pl-0 pr-0">
       {% for item in current_menu.submenu('communities').children if item.permissions == True or permissions[item.permissions] %}
+        {% if item.name == "curation_policy" and not community.metadata.curation_policy %}
```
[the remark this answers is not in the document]
jennur: I will test it! Thanks!

invenio_communities/templates/semantic-ui/invenio_communities/details/header.html
```diff
@@ -140,13 +140,15 @@ <h1 class="ui medium header mb-0">{{ community.metadata.title }}</h1>
     <div
       class="ui container secondary pointing stackable menu page-subheader pl-0 pr-0">
       {% for item in current_menu.submenu('communities').children if item.permissions == True or permissions[item.permissions] %}
+        {% if item.visible %}
```
slint, line 143:
```suggestion
      {% for item in current_menu.submenu('communities').children if item.permissions == True or permissions[item.permissions] and item.visible %}
```

minor/nit: you can add it to the for loop `if ...` clause also to avoid the extra nesting :)

kpsherva (approved):

invenio_communities/views/decorators.py
```diff
@@ -29,6 +29,7 @@ def view(**kwargs):
                 id_=pid_value, identity=g.identity
             )
             kwargs["community"] = community
+            request.community = community.to_dict()
```
ntarocco, line 32: We normally use the `g` context, and not the `request` context. From the Flask doc:
- [g](https://flask.palletsprojects.com/en/2.3.x/api/#flask.g): `This is a good place to store resources during a request.`
- [request](https://flask.palletsprojects.com/en/2.3.x/api/#flask.request): `To access incoming request data, you can use the global request object. Flask parses incoming request data for you and gives you access to it through that global object.`

Minor: `community.to_dict()` is repeated below, it could be set once, and re-used.

Moderate: can you add a comment to explain why? It won't be easy to understand why this was done, and it might break quite easily.

invenio_communities/views/ui.py
```diff
@@ -65,6 +65,26 @@ def _can_create_community():
     return can_create
 
 
+def _has_about_page_content():
+    """Function used to check if about page has content."""
+    community = request.community
```
ntarocco, line 70: See comment above about using `g` instead. Very minor: I would also call it `current_community` to make it very clear.

PR #1004, opened 2023-08-31
components: fix visibility permission check

kpsherva:
* closes https://github.com/zenodo/zenodo-rdm/issues/470

### DIFF
--- a/invenio_communities/communities/services/components.py
+++ b/invenio_communities/communities/services/components.py
@@ -82,7 +82,7 @@ class CommunityAccessComponent(ServiceComponent):
         is_restricting = new_visibility == "restricted"
 
         if (
-            VisibilityEnum(record.access.visibility) != new_visibility
+            VisibilityEnum(record.access.visibility) != VisibilityEnum(new_visibility)
             and record_has_defined_access
         ):
             self.service.require_permission(identity, "manage_access", record=record)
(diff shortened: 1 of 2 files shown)


invenio_communities/communities/services/components.py
```diff
@@ -82,7 +82,7 @@ def _populate_access_and_validate(self, identity, data, record, **kwargs):
         is_restricting = new_visibility == "restricted"
 
         if (
-            VisibilityEnum(record.access.visibility) != new_visibility
+            VisibilityEnum(record.access.visibility).value != new_visibility
```
max-moser, line 85: is `new_visibility` always a string?
```suggestion
            VisibilityEnum(record.access.visibility) != VisibilityEnum(new_visibility)
```
going the other way round may allow `new_visibility` to be either an enum value, or a string.
the casts/constructors may introduce a bit of overhead, but i have a feeling this won't have a huge impact.

ntarocco: Indeed, there is a `.get` above.
The change also looks strange... are we missing a test to catch this?
Can you also check the other occurrences of `VisibilityEnum`?

kpsherva: visibility will always be a string https://github.com/inveniosoftware/invenio-communities/blob/0b1559445919ed76c1d033737554b4c2fd531b43/invenio_communities/communities/schema.py#L79

unfortunately we use the string comparison in most of the places, so it would be a bigger change to find all the occurrences, I would probably do a separate refactoring for that, since it needs time. 
I will add a test though because we were missing one

PR #1023, opened 2023-09-20
User moderation: Delete a user's communities on block

max-moser:
When a user gets blocked, delete all communities of which they are the only owner

### DIFF
--- a/invenio_communities/communities/services/components.py
+++ b/invenio_communities/communities/services/components.py
@@ -237,7 +237,7 @@ class CustomFieldsComponent(ServiceComponent):
 class CommunityDeletionComponent(ServiceComponent):
     """Service component for record deletion."""
 
-    def delete_community(self, identity, data=None, record=None, **kwargs):
+    def delete(self, identity, data=None, record=None, **kwargs):
         """Set the community's deletion status and tombstone information."""
         # Set the record's deletion status and tombstone information
         record.deletion_status = CommunityDeletionStatusEnum.DELETED
@@ -250,23 +250,19 @@ class CommunityDeletionComponent(ServiceComponent):
         """Update the community's tombstone information."""
         record.tombstone = data
 
-    def restore_community(self, identity, data=None, record=None, **kwargs):
+    def restore(self, identity, data=None, record=None, **kwargs):
         """Reset the community's deletion status and tombstone information."""
         record.deletion_status = CommunityDeletionStatusEnum.PUBLISHED
 
         # Remove the tombstone information
         record.tombstone = None
 
-        # Set a record to 'metadata only' if its files got cleaned up
-        if not record.files.entries:
-            record.files.enabled = False
-
-    def mark_community(self, identity, data=None, record=None, **kwargs):
+    def mark(self, identity, data=None, record=None, **kwargs):
         """Mark the community for purge."""
         record.deletion_status = CommunityDeletionStatusEnum.MARKED
         record.tombstone = record.tombstone
 
-    def unmark_community(self, identity, data=None, record=None, **kwargs):
+    def unmark(self, identity, data=None, record=None, **kwargs):
         """Unmark the community for purge, resetting it to soft-deleted state."""
         record.deletion_status = CommunityDeletionStatusEnum.DELETED
         record.tombstone = record.tombstone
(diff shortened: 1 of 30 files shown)


invenio_communities/communities/services/service.py
```diff
@@ -463,18 +448,27 @@ def featured_delete(
     # Deletion workflows
     #
     @unit_of_work()
-    def delete_community(self, identity, id_, data, expand=False, uow=None):
+    def delete_community(
+        self, identity, id_, data=None, revision_id=None, expand=False, uow=None
```
max-moser, line 452: the `revision_id` is accepted **but ignored** because the old signature ([and the resource actually provides a value from the `if_match` request context header](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/resources/records/resource.py#L144))

not sure if that's the way to go?

invenio_communities/communities/services/components.py
```diff
@@ -257,10 +257,6 @@ def restore_community(self, identity, data=None, record=None, **kwargs):
         # Remove the tombstone information
         record.tombstone = None
 
-        # Set a record to 'metadata only' if its files got cleaned up
-        if not record.files.entries:
-            record.files.enabled = False
```
max-moser, line 262: that caused the community's settings page to break after a restore

invenio_communities/communities/services/service.py
```diff
@@ -127,7 +110,9 @@ def search_user_communities(
         # Prepare and execute the search
         params = params or {}
 
-        current_user_filter = CommunityMembers().query_filter(identity)
+        current_user_filter = CommunityMembers().query_filter(identity) & dsl.Q(
+            "term", **{"deletion_status": CommunityDeletionStatusEnum.PUBLISHED.value}
+        )
```
max-moser, line 115: this will certainly hide deleted records from the search results for the user's communities (without it, they'd still pop up)

maybe this should be done via generator query filters somehow?

invenio_communities/requests/user_moderation/actions.py
```diff
@@ -1,26 +1,114 @@
...
+        for m in mem_model_cls.query.filter(mem_model_cls.role == "owner").all()
+    ]:
+        comm_owners[comm_owner.community_id].append(comm_owner)
+
+    # filter for communities that are owned solely by the user in question
+    relevant_comm_ids = [
+        comm_id
+        for comm_id, owners in comm_owners.items()
+        if len(owners) == 1 and str(owners[0].user_id) == user_id
+    ]
```
max-moser, line 51: i have a feeling that this should be doable with a `group_by` and/or `having` for better performance

invenio_communities/communities/services/service.py
```diff
@@ -500,6 +492,13 @@ def delete_community(self, identity, id_, data, expand=False, uow=None):
             expand=expand,
         )
 
+    @unit_of_work()
+    def delete(
+        self, identity, id_, data=None, expand=False, revision_id=None, uow=None
+    ):
+        """Alias for ``delete_community()``."""
+        return self.delete_community(identity, id_, data, expand=expand, uow=uow)
```
zzacharo, line 500: why don't we just keep the `delete` and we introduce a `delete_community`?

invenio_communities/communities/services/service.py
zzacharo: We should [check the revision id as it is done in the RecordService](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L491)

invenio_communities/requests/user_moderation/actions.py
zzacharo: I think we have a helper function to [get all memberships of a user](https://github.com/inveniosoftware/invenio-communities/blob/cfc230e8fb55f9fba8b75b45b92ba01705752f35/invenio_communities/members/records/api.py#L105). It returns though all memberships and not filtered by a specific role but it could be improved. Ping @ntarocco I think in this case the performance issue with getting all communities of a user from DB is done on demand so it should not affect overall performance right?

invenio_communities/communities/services/service.py
zzacharo: PIng @kpsherva

invenio_communities/requests/user_moderation/actions.py
kpsherva: to be discussed, it seems to me that it is not a blocker for now

PR #1033, opened 2023-09-26
default community: add possibility to set to None

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/188

### DIFF
--- a/invenio_communities/records/records/systemfields/communities/manager.py
+++ b/invenio_communities/records/records/systemfields/communities/manager.py
@@ -16,6 +16,7 @@ from invenio_db import db
 from invenio_records.api import Record
 
 from invenio_communities.communities.records.api import Community
+from invenio_communities.errors import SetDefaultCommunityError
 
 
 class CommunitiesRelationManager:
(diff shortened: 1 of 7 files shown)


invenio_communities/records/records/systemfields/communities/manager.py
```diff
@@ -166,7 +166,7 @@ def default(self, community_or_id):
         not, then use ``.add(community, default=True)`` instead.
         """
         id_ = self._to_id(community_or_id)
-        if id_ not in self._communities_ids:
+        if id_ is not None and id_ not in self._communities_ids:
```
jrcastro2, line 169: Question: How can id_ be None? if that happens I would say it's a mistake when calling this setter. In any case, if we want to control this error I would be in favour of raising another exception for this case something like 
```python
raise AttributeError(
                "Cannot set community as the default. "
                "Missing community_id."
            )
```

anikachurilova: We need to add possibility to remove the default community at all. In that case it will be None

jrcastro2: Ah right, I missunderstood it, it makes sense now, thanks :+1:

rekt-hard (approved):

kpsherva (approved):

PR #1043, opened 2023-10-06
header: add manage community button

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/262

image


invenio_communities/permissions.py
```diff
@@ -162,6 +162,8 @@ class CommunityPermissionPolicy(BasePermissionPolicy):
     # correct permissions based on which the field will be exposed only to moderators
     can_moderate = [Disable()]
 
+    can_manage_communities = [UserManager, SystemProcess()]
```
kpsherva, line 165: Possible name clash with `can_manage_access`, maybe `can_administrate_communities` would be better? maybe we can come up together with something more suitable. otherwise LGTM!

PR #1062, opened 2023-10-30
jsonschema: allow none value in description

jrcastro2:
* marshmallow schema remove none values for description
* closes https://github.com/zenodo/rdm-project/issues/489

### DIFF
--- a/invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
+++ b/invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
@@ -60,7 +60,7 @@
         },
         "description": {
           "description": "Community description.",
-          "type": "string"
+          "type": ["string", "null"]
         },
         "type": {
           "description": "Community type.",
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -130,6 +130,11 @@ class CommunityMetadataSchema(Schema):
     # TODO: Add when general vocabularies are ready
     # domains = fields.List(fields.Str())
 
+    @post_dump
+    def remove_none_values(self, data, **kwargs):
+        """Remove none values."""
+        return {key: value for key, value in data.items() if value is not None}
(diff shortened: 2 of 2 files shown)


invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
```diff
@@ -60,7 +60,7 @@
         },
         "description": {
           "description": "Community description.",
-          "type": "string"
+          "type": ["string", "null"]
```
ntarocco, line 63: I would not make this change because of the data in Zenodo, if possible. Otherwise, all the non required scheme should get the same change.

invenio_communities/communities/schema.py
```diff
@@ -130,6 +130,11 @@ class CommunityMetadataSchema(Schema):
     # TODO: Add when general vocabularies are ready
     # domains = fields.List(fields.Str())
 
+    @post_dump
+    def remove_none_values(self, data, **kwargs):
+        """Remove none values."""
+        return {key: value for key, value in data.items() if value is not None}
```
ntarocco, line 136: Here is the classic discussion between: do we allow the client sending `description: None`, and we strip it out, or we fail and we force the client to send a not null value, or omit the field.

jrcastro2: Here we are actually cleaning the data that is dumped from our DB, by default marsmallow doesn't allow None values, to do se we need to enable the ` allow_none=True` (meaning that we don't allow None values at the moment)

ntarocco: Sorry you are right, it is the dump, not the load :)

slint:
After IRL discussion we don't want to proceed with this kind of change. Though we should discuss what is our take on optional/nullable fields: https://github.com/inveniosoftware/invenio-app-rdm/issues/2529

kpsherva:
based on the last comment, closing

PR #1082, opened 2024-02-06
systemfields: add parent community

jrcastro2:
* closes https://github.com/zenodo/zenodo-rdm/issues/696

### DIFF
--- a/invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
+++ b/invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
@@ -145,6 +145,15 @@
         }
       }
     },
+    "parent": {
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "Identifier of the parent community."
+        }
+      }
+    },
     "theme": {
       "type": "object",
       "description": "Community theme settings.",
--- a/invenio_communities/communities/records/systemfields/parent_community.py
+++ b/invenio_communities/communities/records/systemfields/parent_community.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2022 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Community PID slug field."""
+from uuid import UUID
+
+from invenio_records.systemfields import SystemField
+from sqlalchemy.orm.exc import NoResultFound
+
+
+def is_valid_uuid(value):
+    """Check if the provided value is a valid UUID."""
+    try:
+        UUID(str(value))
+        return True
+    except (ValueError, AttributeError, TypeError):
+        return False
+
+
+class ParentCommunityField(SystemField):
+    """System field for parent community."""
+
+    def __init__(self, key="parent"):
+        """Create a new ParentCommunityField instance."""
+        super().__init__(key=key)
+
+    def obj(self, instance):
+        """Get the access object."""
+        obj = self._get_cache(instance)
+        if obj is not None:
+            return obj
+
+        value = self.get_dictkey(instance)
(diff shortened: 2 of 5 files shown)


invenio_communities/communities/records/jsonschemas/communities/communities-v1.0.0.json
```diff
@@ -145,6 +145,10 @@
         }
       }
     },
+    "parent": {
+      "type": "string",
+      "description": "Identifier of the parent community."
+    },
```
slint, line 151:
```suggestion
    "parent": {
      "type": "object",
      "properties": {
        "id": {"type": "keyword"}
      }
    },
```

invenio_communities/communities/records/systemfields/parent_community.py
```diff
@@ -0,0 +1,77 @@
...
+        self._set_cache(instance, obj)
+        return obj
+
+    def set_obj(self, record, obj):
+        """Set the access object."""
+        if is_valid_uuid(obj):
+            try:
+                # Attempt to retrieve the community to confirm its existence
+                current_communities.service.record_cls.get_record(obj)
+                record["parent"] = dict(id=obj)
```
yashlamba, line 54:
```suggestion
                record["parent"] = dict(id=str(obj))
```

Question: Are we waiting some schema changes for this?

This looks like an issue because the type of obj would UUID, which would fail validation.

jrcastro2: Right, I guess we can accept both the string and the UUID, thanks for highlighting this!

invenio_communities/communities/services/service.py
```diff
@@ -20,6 +20,7 @@
     ServiceSchemaWrapper,
 )
 from invenio_records_resources.services.uow import (
+    RecordBulkIndexOp,
```
slint, line 23:
```suggestion
```

nit: unused import?

invenio_communities/communities/records/systemfields/parent_community.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Helper function to validate and set parent community
+        def _validate_and_set_parent(community):
+            if (
+                community.access.visibility_is_public
+                and record.access.visibility_is_public
+            ):
+                record["parent"] = {"id": str(community.id)}
+            else:
+                raise ValueError("Parent community and child community must be public.")
```
slint, line 63: minor: logically this would belong in the service layer, since here (in the data layer) we only care about managing the child/parent relationship in terms of data (not business rules/validation).

invenio_communities/communities/records/systemfields/parent_community.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if is_valid_uuid(obj):
+            try:
+                # Attempt to retrieve the community to confirm its existence
+                parent_community = current_communities.service.record_cls.get_record(
```
slint, line 68: minor: this relies again on "service layer" elements. It would be better to stay in the data layer and e.g.:

```suggestion
                parent_community = record.get_record(obj)
                # or 
                parent_community = Community.get_record(obj)                
```

invenio_communities/communities/records/systemfields/parent_community.py
```diff
@@ -0,0 +1,90 @@
...
+                    obj
+                )
+                _validate_and_set_parent(parent_community)
+            except Exception as e:
```
slint, line 72: minor: is there a more specific exception maybe? like `sqlalchemy.errors.NoResultFound` or similar

slint (commented):
Reviewed with @0einstein0. Some minor comments on mixing service/data layers, but otherwise looks good.

invenio_communities/communities/records/systemfields/parent_community.py
jrcastro2: I have removed it. Pinging @yashlamba as we probably want to add this check in the service layer.

PR #1086, opened 2024-02-07
Add dereferencing parent comm

jrcastro2:
* closes https://github.com/zenodo/zenodo-rdm/issues/696

PR #1088, opened 2024-02-08
service: add update_parent_community service method

yashlamba:
Merge after: #1086

Added service method to add parent to a community.

### DIFF
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -9,7 +9,6 @@
 
 """Community schema."""
 import re
-from copy import deepcopy
 from functools import partial
 from uuid import UUID
 
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -261,7 +261,7 @@ class CommunityParentSchema(BaseCommunitySchema):
 class CommunitySchema(BaseCommunitySchema):
     """Community schema."""
 
-    parent = NestedAttribute(CommunityParentSchema)
+    parent = NestedAttribute(CommunityParentSchema, allow_none=True)
 
     @post_dump
     def post_dump(self, data, many, **kwargs):
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -9,6 +9,7 @@
 
 """Community schema."""
 import re
+from copy import deepcopy
 from functools import partial
 from uuid import UUID
 
@@ -252,16 +253,11 @@ class BaseCommunitySchema(BaseRecordSchema, FieldPermissionsMixin):
 class CommunityParentSchema(BaseCommunitySchema):
     """Community parent schema."""
 
-    id = fields.String(required=True)
-    slug = SanitizedUnicode()
-    metadata = NestedAttribute(CommunityMetadataSchema)
-    access = NestedAttribute(CommunityAccessSchema)
-
 
 class CommunitySchema(BaseCommunitySchema):
     """Community schema."""
 
-    parent = NestedAttribute(CommunityParentSchema, allow_none=True)
+    parent = NestedAttribute(CommunityParentSchema, dump_only=True, allow_none=True)
 
     @post_dump
     def post_dump(self, data, many, **kwargs):
@@ -271,11 +267,15 @@ class CommunitySchema(BaseCommunitySchema):
             data.pop("parent", None)
         return data
 
-    @post_load
-    def filter_parent_id(self, in_data, **kwargs):
+    @post_load(pass_original=True)
+    def filter_parent_id(self, in_data, original_data, **kwargs):
         """Simply keep the parent id."""
-        if in_data.get("parent"):
-            in_data["parent"] = dict(id=in_data.get("parent", {}).get("id"))
+        if "parent" in original_data:
+            in_data["parent"] = (
+                dict(id=original_data["parent"]["id"])
+                if original_data["parent"]
+                else None
+            )
         return in_data
 
     @pre_load
--- a/invenio_communities/communities/schema.py
+++ b/invenio_communities/communities/schema.py
@@ -248,31 +248,20 @@ class BaseCommunitySchema(BaseRecordSchema, FieldPermissionsMixin):
 
         return data
 
-    @pre_load
-    def initialize_custom_fields(self, data, **kwargs):
-        """Ensure custom fields are initialized.
-
-        We need to do that so that validation can take place in case a configured
-        field is marked as required.
-        """
-        data.setdefault("custom_fields", {})
-        return data
-
-    @post_load
-    def lowercase(self, in_data, **kwargs):
-        """Ensure slug is lowercase."""
-        in_data["slug"] = in_data["slug"].lower()
-        return in_data
-
 
 class CommunityParentSchema(BaseCommunitySchema):
     """Community parent schema."""
+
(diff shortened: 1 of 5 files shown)


invenio_communities/communities/services/service.py
```diff
@@ -714,6 +714,30 @@ def on_relation_update(self, identity, record_type, records_info, notif_time):
 
         return True
 
+    @unit_of_work()
+    def update_parent_community(self, identity, id_, parent_id, uow=None):
```
slint, line 718: minor/question: isn't it possible to go via the usual `update` method (similar to how the `theme` is set?). In general we should try to avoid adding new service methods.

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,20 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
+
+    def update_parent(self, identity, data=None, record=None, **kwargs):
+        """Update parent community of a community."""
+        try:
+            parent = self.service.record_cls.pid.resolve(data)
+            if parent.parent:
+                raise ValidationError("Child community cannot be parent.")
+        except PIDDoesNotExistError:
+            raise ValidationError("Parent community does not exist.")
+        record.parent = parent
```
slint, line 314: minor: since we have a dedicated method on the service that "does the thing it say", it would be better to have the logic there (since components are about "side-effects" and isolating their business logic).

If we move to use the `update` service method, this code can stay. We should also handle other methods then, for sure `create`.

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,20 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
 
+class CommunityParentComponent(ServiceComponent):
+    """Service Component for Community parent."""
+
+    def update_parent(self, identity, data=None, record=None, **kwargs):
+        """Update parent community of a community."""
+        try:
+            parent = self.service.record_cls.pid.resolve(data)
+            if parent.parent:
+                raise ValidationError("Child community cannot be parent.")
```
slint, line 311:
```suggestion
                raise ValidationError("Assigned parent community cannot also have a parent.")
```

IIUC, this is to prevent having 2-levels of parents?

slint, line 311: I would also add a check for `child != parent` to avoid some weird recursive serialization 😅

slint (commented):
Logic looks good, some minor comments on placement and some additional validation

invenio_communities/communities/services/components.py
yashlamba: Yes

invenio_communities/communities/schema.py
```diff
@@ -240,31 +248,20 @@ def post_dump(self, data, many, **kwargs):
...
-        return in_data
-
 
 class CommunityParentSchema(BaseCommunitySchema):
     """Community parent schema."""
 
+    id = fields.String(required=True)
+    slug = SanitizedUnicode()
+    metadata = NestedAttribute(CommunityMetadataSchema)
+    access = NestedAttribute(CommunityAccessSchema)
```
yashlamba, line 258: @slint need your review on this.

invenio_communities/communities/schema.py
```diff
@@ -240,31 +248,20 @@ def post_dump(self, data, many, **kwargs):
...
+
 
 class CommunitySchema(BaseCommunitySchema):
     """Community schema."""
 
-    parent = NestedAttribute(CommunityParentSchema, dump_only=True)
+    parent = NestedAttribute(CommunityParentSchema)
```
jrcastro2, line 264: Minor: to be able to set it to None

```suggestion
    parent = NestedAttribute(CommunityParentSchema, allow_none=True)
```

invenio_communities/communities/schema.py
slint: But can't we make the `parent` field dump-only, and then using the `post_load` fish out the `parent.id` if necessary? Then there's no need to align e.g. specific fields that we load/dump.

jrcastro2: Hmm, if I am not mistaken, if we mark the parent as `dump_only` inside the `post_load` hook the parent field is empty , maybe @yashlamba can confirm this.

From what I see in the docs (see image below) we could do what you say but we would need to set the `unknown` option to `INCLUDE` (so that it's not empty on `post_load` and we can fish the id)
Screenshot from 2024-02-15 10-14-24

slint: `@post_load` can have access to the original data by passing `pass_original=True`, i.e.:


```python
    @post_load(pass_original=True)
    def load_parnet_id(self, data, original_data=None, **kwargs):
        """Load parent ID if present."""
        parent_id = original_data.get("parent", {}).get("id")
        if parent_id:
            data["parent"] = {"id": parent_id}
        return data
```

jrcastro2: Ah cool! that's way better :+1:

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,59 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
+        try:
+            parent = self.service.record_cls.pid.resolve(parent_data["id"])
+            if parent.parent:
+                raise ValidationError(
+                    "Assigned parent community cannot also have a parent."
+                )
+            elif child.id == parent.id:
+                raise ValidationError(
+                    "Assigned parent community cannot be the same as child."
+                )
```
yashlamba, line 319: Add check for `children.allow`

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,59 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
+        except PIDDoesNotExistError:
+            raise ValidationError("Assigned parent community does not exist.")
+        return parent
+
+    def create(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed theme to the record."""
+        if "parent" in data:
+            self.service.require_permission(identity, "update_parent", record=record)
```
yashlamba, line 327:
```suggestion
            self.service.require_permission(identity, "manage_parent", record=record)
```

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,59 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
+            parent = self._validate_and_get_parent(data["parent"], record)
+            record.parent = parent
+
+    def update(self, identity, data=None, record=None, **kwargs):
+        """Update parent community of a community."""
+        if "parent" in data:
+            self.service.require_permission(identity, "update_parent", record=record)
```
yashlamba, line 334:
```suggestion
            self.service.require_permission(identity, "manage_parent", record=record)
```

invenio_communities/communities/services/components.py
```diff
@@ -311,4 +364,6 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
     FeaturedCommunityComponent,
     OAISetComponent,
     CommunityDeletionComponent,
+    ChildrenComponent,
+    CommunityParentComponent,
```
yashlamba, line 368: Update in invenio-rdm-records

yashlamba (commented):
Reviewed with @zzacharo

slint (approved):

PR #1091, opened 2024-02-12
parent_community: use OS instead of DB queries for search result

jrcastro2:
* updates schema to laod parent.id only
* closes https://github.com/zenodo/zenodo-rdm/issues/700

### DIFF
--- a/invenio_communities/communities/records/systemfields/children.py
+++ b/invenio_communities/communities/records/systemfields/children.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2023 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Children system field."""
+
+from invenio_records.systemfields import SystemField
+
+
+class Children:
+    """Children object."""
+
+    def __init__(self, allow=None):
+        """Create a Children object."""
+        self._allow = allow or False
+
+    @property
+    def allow(self):
+        """Get the allow."""
+        return self._allow
+
+    @allow.setter
+    def allow(self, value):
+        """Set the allow."""
+        if not isinstance(value, bool):
+            raise ValueError("Invalid value for allow, it must be a boolean.")
+        self._allow = value
+
+    @classmethod
+    def from_dict(cls, data):
+        """Create a Children object from a dictionary."""
+        return cls(allow=data.get("allow"))
+
+    def dump(self):
+        """Dump the Children object to a dictionary."""
+        return {"allow": self.allow}
+
+
+class ChildrenField(SystemField):
+    """System field for children of a community."""
+
+    children_obj_class = Children
+
+    def __init__(self, key="children", children_obj_class=None):
+        """Create a new ChildrenField instance."""
+        self._children_obj_class = children_obj_class or self.children_obj_class
+        super().__init__(key=key)
+
+    def obj(self, instance):
+        """Get the Children object."""
+        obj = self._get_cache(instance)
+        if obj is not None:
+            return obj
+
+        data = self.get_dictkey(instance)
+        if data:
+            obj = self._children_obj_class.from_dict(data)
+        else:
+            obj = self._children_obj_class()
+
+        self._set_cache(instance, obj)
+        return obj
+
+    def set_obj(self, record, obj):
+        """Set the Children object."""
+        # We accept both dicts and Children class objects.
+        if isinstance(obj, dict):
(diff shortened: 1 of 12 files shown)


invenio_communities/communities/schema.py
```diff
@@ -265,14 +265,25 @@ class CommunitySchema(BaseCommunitySchema):
...
         data = super().post_dump(data, many, **kwargs)
         if data.get("parent") is None:
             data.pop("parent", None)
         return data
+    
+    @post_load
+    def filter_parent_id(self, in_data, **kwargs):
+        """Simply keep the parent id."""
+        if in_data.get("parent"):
+            in_data["parent"] = dict(id=in_data.get("parent", {}).get("id"))
```
slint, line 285: minor: there are two input cases we have to handle:

- `{"parent": {"id": "<some_id>"}` -  we filter and just keep the `parent.id` key
- `{"parent": None}` - We keep the data as is so that we can "unset" the community

invenio_communities/communities/schema.py
```diff
@@ -265,14 +265,25 @@ class CommunitySchema(BaseCommunitySchema):
     """Community schema."""
 
     parent = NestedAttribute(CommunityParentSchema, dump_only=True)
-
+    # When loading the parent community we don't require any fields as we will only use the ID
+    id = fields.String(dump_only=False)
+    metadata = NestedAttribute(CommunityMetadataSchema, required=False) 
+    access = NestedAttribute(CommunityAccessSchema, required=False)
```
slint, line 271: minor: it's actually ok to pass all the fields even when just updating the parent, so we can remove these. This is so that we keep the `PUT` semantics, which requires the entire representation (we do divert though for `theme` and `parent`, where we have some `PATCH` behavior).

invenio_communities/communities/records/systemfields/parent_community.py
```diff
@@ -51,6 +51,7 @@ def set_obj(self, record, obj):
         # Check if obj is None and remove 'parent' key from record
         if obj is None:
             record.pop("parent", None)
+            self._set_cache(record, None)
```
slint, line 54: Nice! I suspect that might be the case in other system fields that don't clear their cache... Can you grep-around maybe to see if it's true?

slint (commented):
Some comments on handling all cases, but otherwise looks good to go soon.

invenio_communities/communities/schema.py
jrcastro2: Right, sorry this is a leftover, I will remove it

invenio_communities/communities/schema.py
jrcastro2: I think that @yashlamba is working on the service method that allows to set/update the parent, maybe you can improve it as you can test it

invenio_communities/communities/records/systemfields/children.py
```diff
@@ -0,0 +1,97 @@
...
+        self._set_cache(instance, obj)
+        return obj
+
+    def set_obj(self, record, obj):
+        """Set the Children object."""
+        # We accept both dicts and Children class objects.
+        if isinstance(obj, dict):
+            obj = self.children_obj_class.from_dict(obj)
+
+        assert isinstance(obj, self.children_obj_class)
```
slint, line 73: minor: same as above, raising a `ValueError` is a bit cleaner

invenio_communities/communities/records/systemfields/children.py
```diff
@@ -0,0 +1,97 @@
...
+            return self
+
+        # access by object
+        return self.obj(record)
+
+    def __set__(self, record, obj):
+        """Set the records Children object."""
+        self.set_obj(record, obj)
+
+    def pre_commit(self, record):
```
slint, line 91: comment/shelve: we should have probably done the same with the parent field, i.e. have a `pre_commit` to manage how/when the field is persisted in the record

invenio_communities/communities/services/components.py
```diff
@@ -300,6 +300,23 @@ def update(self, identity, data=None, record=None, errors=None, **kwargs):
...
+        # We check if the children field is passed and it's different to the values stored
+        if "children" in data and data["children"] != record.get("children", {}):
+            self.service.require_permission(identity, "manage_children", record=record)
+            record["children"].update(data["children"])
+
+    def create(self, identity, data=None, record=None, **kwargs):
+        """Create handler."""
+        if "children" in data:
+            self.service.require_permission(identity, "manage_children", record=record)
+            record["children"] = data["children"]
```
slint, line 317: nit: to preserve the logical/original order this method should come before `update`

slint (approved):
LGTM! Some minor comments and maybe a bigger issue to shelve and apply the same structure/logic to the parent field.

PR #1095, opened 2024-02-14
views: fix menu order

kpsherva:


invenio_communities/communities/records/api.py
```diff
@@ -74,8 +74,6 @@ class Community(Record):
     custom_fields = DictField(clear_none=True, create_if_missing=True)
     theme = DictField(clear_none=True)
 
-    theme = DictField(clear_none=True)
```
slint, line 77: minor: where theme?

0einstein0: in line 75 above, it was duplicated

PR #1112, opened 2024-02-27
services: added bulk 'add to parent' call

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1677

### DIFF
--- a/invenio_communities/communities/services/service.py
+++ b/invenio_communities/communities/services/service.py
@@ -714,6 +714,25 @@ class CommunityService(RecordService):
 
         return True
 
+    @unit_of_work()
+    def bulk_update_parent(self, identity, community_ids, parent_id, uow=None):
+        """Bulk updates communities with a new parent.
+
+        :param identity: The identity of the user performing the action.
+        :param community_ids: The list of community IDs to add to the new parent.
+        :param parent_id: The ID of the new parent community.
+        """
+        parent = self.read(identity, parent_id)
+        for comm_id in community_ids:
+            child = self.read(identity, comm_id)
+            self.update(
+                identity,
+                comm_id,
+                data={**child.data, "parent": {"id": parent.id}},
+                uow=uow,
+            )
+        return True
+
 
 @cached_with_expiration
 def get_cached_community_slug(
--- a/tests/communities/test_services.py
+++ b/tests/communities/test_services.py
@@ -730,3 +730,35 @@ def test_parent_update_child_children_are_allowed(community_service, comm):
             id_=str(child.id),
             data={**child.data, "parent": {"id": str(parent.id)}},
         )
+
+
+def test_bulk_update_parent(
+    community_service, parent_community, comm, restricted_community
+):
+    """Test bulk add parent to children."""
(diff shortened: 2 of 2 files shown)


invenio_communities/communities/services/service.py
```diff
@@ -714,6 +714,25 @@ def on_relation_update(self, identity, record_type, records_info, notif_time):
...
+        :param comm_slugs: The list of community slugs to add to the new parent.
+        :param new_parent: The slug of the new parent community.
+        """
+        t_comm = self.read(identity, new_parent)
+        for c_slug in comm_slugs:
+            c_comm = self.read(identity, c_slug)
+            self.update(
+                identity,
+                c_slug,
+                data={**c_comm.data, "parent": {"id": t_comm.id}},
```
jrcastro2, line 731: Question: Should there be a check for an already existing parent, or is it okay to replace it? (maybe adding a force option to replace would be useful (?))

To us it looks like if we want to allow updates at least we should call it `bulk_update_parent` instead.

invenio_communities/communities/services/service.py
```diff
@@ -714,6 +714,25 @@ def on_relation_update(self, identity, record_type, records_info, notif_time):
 
         return True
 
+    @unit_of_work()
+    def bulk_add_to_parent(self, identity, comm_slugs, new_parent, uow=None):
```
jrcastro2, line 718:
```suggestion
    def bulk_add_parent(self, identity, comm_slugs, new_parent, uow=None):
```

jrcastro2 (commented):
Pair-reviewed with @0einstein0

invenio_communities/communities/services/service.py
alejandromumo: I disagree on this one, semantically `community_service.bulk_add_parent` seems it's adding parents in bulk. We're adding records, in bulk, to one parent (`community_service.bulk_add_to_parent)`. But I'm open to other suggestions.

invenio_communities/communities/services/service.py
alejandromumo: Regarding the questions:

1) will be added, I don't see why not. 
2) I also agree with this one. The way I saw it, we needed a way to "transfer" multiple communities to a new parent. Thus, the wording `add to parent`. But you're right, this is updating the parent (not adding).

invenio_communities/communities/services/service.py
```diff
@@ -714,6 +714,25 @@ def on_relation_update(self, identity, record_type, records_info, notif_time):
 
         return True
 
+    @unit_of_work()
+    def bulk_update_parent(self, identity, comm_slugs, new_parent, uow=None):
```
slint, line 718:
```suggestion
    def bulk_update_parent(self, identity, community_ids, parent, uow=None):
```

minor: slugs and UUIDs are resolved uniformly in communities, so it's ok to be "generic" about it :)

Also some renaming to simplify.

slint (commented):
LGTM, minor namings nits (I will apply and merge)

Issue #1108, opened 2024-02-26
Remove "Members" menu item (new Members visibility setting)


lnielsen:
The ``Members`` menu item in the community should only be visible to members of the community.

- [ ] Implement a new setting "Members visibility"  with support for setting it through the API, similar to "Community visibility". Default to ``public``.
- [ ] Add UI to Privileges page (``https://sandbox.zenodo.org/communities/:id/settings/privileges``)

![Image](https://github.com/inveniosoftware/invenio-communities/assets/1698163/4a08182d-f22d-4c04-aed2-95fe19d4c901)

- Please use exact text and icons as in screenshot.
-  Increase font-size of headline for both Community and Member visibility

PR #1116, opened 2024-03-04
global: implement access.members_visibility field

slint:
- Closes https://github.com/inveniosoftware/invenio-communities/issues/1108


slint:
Merging since we're on a pinch (but will keep in the review column for any fixes).

PR #1124, opened 2024-04-08
models: fix memberships querying

slint:
* Fixes an edge-case where if no filtering parameters are passed all
  community/role pairs are returned.


zzacharo (approved):

slint:
Updated the relevant test, and also confirmed that without the fix commit it fails 🚀

PR #1157, opened 2024-06-21
Subcommunities: fix request redirect url after new request (UI)

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/882

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/subcommunity/new.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/subcommunity/new.js
@@ -110,13 +110,16 @@ class CommunityCreateForm extends Component {
     const { hasCommunity } = this.state;
     const { communityId } = this.props;
     let payload = {};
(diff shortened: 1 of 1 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/subcommunity/new.js
```diff
@@ -124,7 +124,7 @@ class CommunityCreateForm extends Component {
     try {
       const response = await this.cancellableCreate.promise;
       setSubmitting(false);
-      window.location.href = response.data.links.settings_html;
+      window.location.href = response.data.links.self_html;
```
alejandromumo, line 127: `self_html` returns the request URL (`/requests/<uuid>`) but it doesn't work. The URL that works is `/communities/eu/requests/<uuid>`. 

I am not sure why this happens but I am investigating it

slint (approved):

PR #1164, opened 2024-06-25
request: added subcommunity type as a function to entry point.

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/891

**For reviewers**

this depends on https://github.com/inveniosoftware/invenio-requests/pull/388 , so checks are failing due to that


invenio_communities/subcommunities/services/service.py
```diff
@@ -51,6 +52,19 @@ def expandable_fields(self):
         """Get expandable fields."""
         return []
 
+    def _is_owner_of(self, identity, community):
+        """Check if the identity provides community ownership."""
+        return next(
```
alejandromumo, line 57: I considered the following ways of checking for ownership:

1. inspecting the identity (current implementation)
2. checking the memberships of the user and look for one that is `owner`

If you have any other ideas, please share

slint (approved):

PR #1174, opened 2024-07-23
fix is verified

utnapischtim:
- fix(verified): group not handled

# NOTES
- [commit](https://github.com/inveniosoftware/invenio-communities/pull/1174/commits/c17a20455bbbbf821ba5677c0ff198fc7f0c3327) fixes https://github.com/inveniosoftware/invenio-app-rdm/issues/2639

### DIFF
--- a/invenio_communities/communities/records/systemfields/is_verified.py
+++ b/invenio_communities/communities/records/systemfields/is_verified.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023-2024 CERN.
+# Copyright (C) 2024 Graz University of Technology.
 #
 # Invenio-Communities is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -23,11 +24,12 @@ class IsVerifiedField(CalculatedIndexedField):
         # import here due to circular import
         from invenio_communities.members.records.api import Member
 
-        community_verified = False
         owners = [m.dumps() for m in Member.get_members(record.id) if m.role == "owner"]
+
+        # community is considered verified if at least one owner is verified
+        # a user is verified if verified_at is not None
         for owner in owners:
-            # community is considered verified if at least one owner is verified
-            if owner["user"]["verified_at"] is not None:
-                community_verified = True
-                break
-        return community_verified
+            if "user" in owner and owner["user"]["verified_at"] is not None:
+                return True
+
+        return False


invenio_communities/communities/records/systemfields/is_verified.py
```diff
@@ -23,11 +24,15 @@ def calculate(self, record):
...
+        # a group is per definition verified, because it could only be created by admin
         for owner in owners:
-            # community is considered verified if at least one owner is verified
-            if owner["user"]["verified_at"] is not None:
-                community_verified = True
-                break
-        return community_verified
+            if "user" in owner and owner["user"]["verified_at"] is not None:
+                return True
+            if "group" in owner:
```
zzacharo, line 35: We discussed with @kpsherva and @slint that maybe for now is better not to take into account groups as part of the verification mechanism. So, just fixing the "user" key access should be enough.

utnapischtim: may i asking why is it better?

that would mean that a community is not verified if a group is the owner.

and yes technically i am fixing two problems with this commit.

zzacharo: Groups at the moment are not part of "user moderation" which means that potentially a user can trick the system (consider the case of Zenodo where users that are potential spammers) and verify their spam community. 

What is the second problem that you are fixing?

utnapischtim: trick the system would only work if the user is part of that group and therefore verified already or?

i mean that the commit fixes the KeyError "user" not existing and that groups are not handled. they are independent of each other so this commit fixes two problems which is not fully mentioned in the commit message and also it not really asked in the issue

zzacharo: > trick the system would only work if the user is part of that group and therefore verified already or?
Yes you are right about it if you always consider that groups are considered "verified". I would feel probably safer to not take into account the groups as they are not part of any moderation and thus is not a conscious choice. 

An alternative would be to have a configurable choice to indicate "trust or not" groups. Maybe  @slint @kpsherva can also share their thoughts :)

kpsherva: I think more prudent approach for now would be not to take groups for granted - I wouldn't mark them as verified by default, by doing that we open a possible "workaround" for verifying the community, but also we are defining a base workflow which might be difficult to change: if we decide to verify groups in any way, the workflow will change and it will be hard to make it backwards compatible if we decide to become less permissive in the verification process in the future - therefore I would also return False here instead

utnapischtim:
should i add the other two commits about fixing the ValidationError handling? i am asking because it would need to update the tests too and not to make work which is not wanted i would like to ask first.

utnapischtim:
> Can you elaborate on the current bug that the `ValidationError` exception fixes? We never caught validation errors properly?

it is from the usability point of view but to fix it properly it produces more work then i thought . the point for me was that if i would remove the owner without stating the group as the only owner i get a validation error back. the visualization could be improved. my simple improvement was to catch the ValidationError and give back only the string. this would show only the message to the user which in my point of view improves the usability. the line "ValidationError" is mostly to technical for users.

but i think we should put a little bit more thoughts into the error handling here and therefore i will remove those two commits so this PR is only on fixing the is verified bug.

zzacharo (approved):

PR #1204, opened 2024-09-05
user_moderation: user search for fetching user communities

slint:
* Uses search results to determine a blocked/restored user's
  communities.
* Uses TaskOp and Unit of Work for sending off action tasks.

### DIFF
--- a/invenio_communities/requests/user_moderation/actions.py
+++ b/invenio_communities/requests/user_moderation/actions.py
@@ -7,60 +7,39 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 """Communities user moderation actions."""
 
-from collections import defaultdict
-
 from invenio_access.permissions import Identity, system_identity
+from invenio_db import db
 from invenio_i18n import lazy_gettext as _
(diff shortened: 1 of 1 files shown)


invenio_communities/requests/user_moderation/actions.py
```diff
@@ -7,60 +7,39 @@
...
+        )
+        return (row[0] for row in query.yield_per(1000))
+    else:
+        search = (
+            RecordsSearchV2(index=mem_cls.index._name)
+            .filter("term", user_id=user_id)
+            .filter("term", role="owner")
+            .source(["community_id"])
+        )
+        return (hit["community_id"] for hit in search.scan())
```
jrcastro2, line 42: Major: Hmm I think we are missing the filtering for the communities that are owned solely by the user, unless I am missing something?

slint: You're right, this doesn't check anymore if the user is the only owner. Part of the change was to improve on the overall performance (before there was some in-memory overfetching). I think adding the check back is not that big of a performance hit, so I'll have a go at it.

Issue #1209, opened 2024-09-20
Exclude 'new' from available community ids / slugs


sakshamarora1:
Creating a community with the identifier as `new` shouldn't be allowed as the URL is already being used for new community creation page.

PR #1210, opened 2024-09-24
communities: config: Update static page URLs to prevent name collisions

sakshamarora1:
Fixes: https://github.com/inveniosoftware/invenio-communities/issues/1209

### DIFF
--- a/invenio_communities/config.py
+++ b/invenio_communities/config.py
@@ -21,8 +21,10 @@ from invenio_communities.communities.services import facets
 
 COMMUNITIES_ROUTES = {
     "frontpage": "/communities",
-    "search": "/communities/search",
-    "new": "/communities/new",
+    "search": "/communities-search",
+    "deprecated_search": "/communities/search",
+    "new": "/communities-new",
+    "deprecated_new": "/communities/new",
     "upload": "/communities/<pid_value>/upload",
     "settings": "/communities/<pid_value>/settings",
     "requests": "/communities/<pid_value>/requests",
--- a/invenio_communities/views/ui.py
+++ b/invenio_communities/views/ui.py
@@ -43,6 +43,7 @@ from .communities import (
     invitations,
     members,
 )
+from .decorators import warn_deprecation
 
 
 #
@@ -150,8 +151,25 @@ def create_ui_blueprint(app):
     )
 
     blueprint.add_url_rule(
+        routes["deprecated_search"],
+        endpoint="deprecated_communities_search",
+        view_func=warn_deprecation(routes["deprecated_search"], routes["search"])(
+            communities_search
+        ),
+        strict_slashes=False,
+    )
+
+    blueprint.add_url_rule(
         routes["new"],
         view_func=communities_new,
+    )
+
+    blueprint.add_url_rule(
+        routes["deprecated_new"],
+        endpoint="deprecated_communities_new",
+        view_func=warn_deprecation(routes["deprecated_new"], routes["new"])(
+            communities_new
+        ),
     )
 
     blueprint.add_url_rule(
--- a/invenio_communities/templates/semantic-ui/invenio_communities/search.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/search.html
@@ -22,7 +22,7 @@
     <h1 class="ui large header m-0">{{_("Communities")}}</h1>
 
     {% if permissions.can_create %}
-      <a class="ui button positive left labeled icon m-0" href="/communities/new">
+      <a class="ui button positive left labeled icon m-0" href="{{ config.COMMUNITIES_ROUTES['new'] }}">
         <i class="plus icon" aria-hidden="true"></i>
         {{ _('New community') }}
       </a>
--- a/invenio_communities/views/decorators.py
+++ b/invenio_communities/views/decorators.py
@@ -9,6 +9,7 @@
 """Decorators."""
 
 from functools import wraps
+from warnings import warn
 
 from flask import g, request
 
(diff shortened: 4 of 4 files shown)


invenio_communities/communities/schema.py
```diff
@@ -76,6 +76,13 @@ def is_not_uuid(value):
         pass
 
 
+def is_not_reserved(value):
+    """Make sure value is not a reserved keyword."""
+    return validate.NoneOf(
+        iterable=["new", "search"], error=("The ID cannot be a reserved keyword.")
```
zzacharo, line 82: question: couldn't we parse the communities routes config to dynamically identify these?

invenio_communities/communities/schema.py
```diff
@@ -76,6 +76,13 @@ def is_not_uuid(value):
         pass
 
 
+def is_not_reserved(value):
+    """Make sure value is not a reserved keyword."""
+    return validate.NoneOf(
+        iterable=["new"], error=("The ID cannot be a reserved keyword.")
```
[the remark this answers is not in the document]
slint: I wouldn't go about changing URLs, since these are often externally referenced in project pages or reports.

As mentioned in https://github.com/inveniosoftware/invenio-communities/pull/1210#discussion_r1774658800, having this being dynamically computed (and then cached) from the registered Flask routes would be the most future-proof.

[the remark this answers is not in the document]
kpsherva: this won't work if the community with slug `featured` is first created and then we want to create a route `communities/featured`. I think we should change the root path

invenio_communities/config.py
```diff
@@ -21,8 +21,8 @@
 
 COMMUNITIES_ROUTES = {
     "frontpage": "/communities",
-    "search": "/communities/search",
-    "new": "/communities/new",
+    "search": "/search/communities/all",
```
ntarocco, line 24: What is the impact of these changes?

invenio_communities/communities/schema.py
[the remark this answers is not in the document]
carlinmack: I think this approach of changing the URL structure is very radical and should be well considered (as URLs should last [2 years, 20 years, even 200 years](https://www.w3.org/Provider/Style/URI)). 

For a user is it not confusing that to create a record you go to `/uploads/new` but `/new/communities/create`? Or is the idea that we going to change all URLs so that they are verb first?

I don't think it's unreasonable to introduce a controlled vocabulary of names that communities aren't allowed, as Github and all social medias do this and so people are understanding. Perhaps when we create a new release of Invenio, in the rare case when we add a reserved keyword, it could be part of the migration instructions to tell maintainers to ask to change the name of the community or disable the feature.

The only reason I'm against the verb first URLs is that it is very easy to create URLs with subpaths that 404. Users of library systems will expect to be able to modify the URL to where they want to go and it would be confusing if `/featured` or '/new/communities` 404s. If we're willing to do the extra work to make sure the URLs are consistent and don't 404, then I think switching to verb first makes sense

invenio_communities/config.py
carlinmack: added my thoughts here https://github.com/inveniosoftware/invenio-communities/pull/1210#discussion_r1784500810

kpsherva: @sakshamarora1 did you check if we don't hardcode these URLs somewhere in the UI?

sakshamarora1: There are, I will update. But I am still not sure if we go ahead and with which change. Waiting for more people to give their opinions.

slint: Given that the impact and mitigations (besides internally changing links in the codebase) are not easy to evaluate (e.g. existing public docs for RDM instances, having to communicate this in release notes, providing a transition, etc.) and can't be fixed similar to the `/communities/<slug> -> /communities/<slug>/records` URL redirect we did for the community records, I'm leaning toward the reserved keywords option.

That can be either done by manually maintaining the list of keywords or automatically calculating it from the registered routes.

kpsherva:
latest proposal (by @slint) on how to handle the links 

```
/communities/new ---> /communities-new
/communities-search ----> /communities
/communities/<slug> -> leave as is
/communities ---> /communities-home
```
and we don't "reserve" any slugs
ping @ntarocco, is that OK?

sakshamarora1:
Reviewed with Nico, going forward with the updated comment.

zzacharo (approved):

invenio_communities/config.py
```diff
@@ -21,8 +21,10 @@
 
 COMMUNITIES_ROUTES = {
     "frontpage": "/communities",
-    "search": "/communities/search",
-    "new": "/communities/new",
+    "search": "/communities-search",
+    "deprecated_search": "/communities/search",
+    "new": "/communities-new",
+    "deprecated_new": "/communities/new",
```
slint, line 27: minor: I think just adding them doesn't automatically register the routes as well, they have to be registered inside the `invenio_communities/views/ui::create_ui_blueprint` function.

invenio_communities/views/ui.py
```diff
@@ -149,11 +149,22 @@ def create_ui_blueprint(app):
...
+    )
+
     blueprint.add_url_rule(
         routes["new"],
         view_func=communities_new,
     )
 
+    blueprint.add_url_rule(
+        routes["deprecated_new"],
+        view_func=communities_new,
```
zzacharo, line 165: How do we communicate the deprecation? I assume it will be part of the release notes but does it make sense to log a deprecation warning in the view method so instance administrator see it if there are still occurrences in their code? ping @slint @ntarocco

slint: It should go in the release notes. Agree, logging a warning is probably the best for visibility, though there should also be the possibility to do something about it (e.g. suggest that if the instance wants to keep the endpoint they should handle a redirect on their side).

invenio_communities/views/communities.py
```diff
@@ -267,6 +268,11 @@ def load_custom_fields(dump_only_required=False):
 @login_required
 def communities_new():
     """Communities creation page."""
+    warn(
```
slint, line 271: minor: shouldn't this only be logged if we know for sure the `/communities/new` endpoint is being used? Otherwise it will show up always, even when folks access the new `/communities-new` endpoint.

sakshamarora1: Yes ideally that should be the case, but having a general warning should be fine since we're gonna remove it in the future anyways?
If we want it only for the deprecated, would it make sense to have a generic wrapper?

slint: My take on warnings is that someone should be able to do something about to make them go away. Imagine e.g. folks that use the `-W` flag in pytest, they wouldn't have any way to address the warning (besides explicitly ignoring it in test config).

In this case it's just going to be noise in the terminal everytime you run the application and access the page (even with the correct URL).

Having a new view function (using a wrapper or similar) and registering would be the way to go IMO.

slint (approved):
LGTM! Thanks for all the work :)

PR #1216, opened 2024-10-01
bug: fix error getting whether children allowed

carlinmack:


invenio_communities/communities/services/config.py
```diff
@@ -72,6 +74,17 @@ class SearchOptions(SearchOptionsBase, SearchOptionsMixin):
...
+    """Determine if children are allowed."""
+    if isinstance(record, dict):
+        return record.get("children", {}).get("allow", False)
+    elif isinstance(record, SimpleNamespace):
+        if hasattr(record, "children"):
+            return getattr(record.children, "allow", False)
+
+    return False
+
+
```
alejandromumo, line 87: I think we have to fix the root of the problem:

https://github.com/inveniosoftware/invenio-communities/blob/f56b388de22c7470dd741ca713b644f3cb137c03/invenio_communities/communities/entity_resolvers.py#L38-L41

```python
    fake_community_obj = SimpleNamespace(
        id=community_dict["id"],
        slug=community_dict["slug"],
       children=community_dict.get("children", {})
    )
```
(please confirm whether `community_dict` has `children` in it, it might be it has to be added somewhere)

...

and then safely get the `children.allow` property:

```python
return getattr(community.children, "allow", False)
```

alejandromumo (approved):

ptamarit (approved):

PR #1223, opened 2024-10-09
subcommunities: updated page layout

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/1006

### DIFF
--- a/invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html
+++ b/invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html
@@ -1,10 +1,10 @@
 {# -*- coding: utf-8 -*-
 
-    This file is part of Invenio.
-    Copyright (C) 2024 CERN.
+This file is part of Invenio.
+Copyright (C) 2024 CERN.
 
-    Invenio is free software; you can redistribute it and/or modify it
-    under the terms of the MIT License; see LICENSE file for more details.
+Invenio is free software; you can redistribute it and/or modify it
+under the terms of the MIT License; see LICENSE file for more details.
 #}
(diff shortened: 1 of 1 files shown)


invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html
```diff
@@ -14,20 +14,34 @@
...
+<div class="ui container fluid page-subheader rel p-10 mr-0-mobile ml-0-mobile">
+  <div class="collection-header ui container align-items-center justify-space-between">
+    <h1 class="ui medium header m-0">{{ config.get("APP_RDM_SUBCOMMUNITIES_LABEL", _("Subcommunities")) }}</h1>
+  </div>
+</div>
+<!-- Breadcrumbs -->
+<div class="ui container mb-0">
+  <div class="ui breadcrumb">
+    <a class="section"
+      href="{{ url_for('invenio_app_rdm_communities.communities_browse', pid_value=community.slug ) }}">
```
slint, line 32: minor: I guess we're aware of the "backwards" reference to Invenio-App-RDM? In theory "Browse" as a concept only exists there.

slint (commented):
LGTM, with an acceptable reverse "dependency" that we can sort out in the future.

invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html
alejandromumo: Yes, this was discussed when we first implemented the browse page. Unfortunately, there's some entanglement that IMO should be cleaned up when we have more time.

PR #1208, opened 2024-09-19
feat: add Subcommunities page

carlinmack:
:heart: Thank you for your contribution!

Close https://github.com/zenodo/zenodo-rdm/issues/990

### Description

> Please describe briefly your pull request.
image

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

PR #1226, opened 2024-10-14
subcommunities: revert resource config change

alejandromumo:
The header must be supported for subcommunity requests to work, otherwise the request fails due to unsupported accept header type. 

I am reverting it to the state that was before these [changes](https://github.com/inveniosoftware/invenio-communities/pull/1208/files).

### DIFF
--- a/invenio_communities/subcommunities/resources/config.py
+++ b/invenio_communities/subcommunities/resources/config.py
@@ -55,6 +55,7 @@ class SubCommunityResourceConfig(ConfiguratorMixin, ResourceConfig):
     # Response handling
     response_handlers = {
         "application/json": json_response_handler,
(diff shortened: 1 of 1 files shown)


carlinmack:
The vendor/UI handler was added so that it would be compatible with the UI for the browse gallery and the subcoms reactsearchkit. Would be interested how you got them to work with the JSON endpoint

alejandromumo:
> The vendor/UI handler was added so that it would be compatible with the UI for the browse gallery and the subcoms reactsearchkit. Would be interested how you got them to work with the JSON endpoint

The subcommunities endpoint was  moved under `communities` resource, this resource only deals with requests now

carlinmack (approved):

PR #1246, opened 2024-11-12
remove flask upper pin

utnapischtim:
- global: make forward compatible to sqlalchemy >= 2
- fix: for compatibility with sqlalchemy >= 2


utnapischtim:
seems not necessary for flask>=3.0

PR #1258, opened 2024-12-06
Allow custom routes to be picked up by community links template

monotasker:

### Description

Currently changes to routes via the COMMUNITIES_ROUTES config variable do not affect the urls in the "links" property of community record results. This PR offers a fix which allows the community links template to pick up custom routes from the config variable.

There are two challenges to solving this issue:
1.  The config variables are not available to the CommunityServiceConfig when it is being initialized
2. The COMMUNITIES_ROUTES url templates use a different templating format than the links template does

I've attempted to solve #1 by delaying the actual creation of the link strings until after the app has been initialized and the config variables are available. Rather than building the link strings when CommunityServiceConfig is initialized, I move the link string construction into a factory class. That factory class is then called during the dumping of records to result objects, and since the `current_app` is available at that point, we can draw in the required COMMUNITIES_ROUTES variables.

I also include in the factory class some string conversion logic that converts the routes strings to the templating format used by the links template.

This may not be the ideal solution to this problem in the long run, but it does solve the problem and does so while allowing the COMMUNITIES_ROUTES to be a single source of truth for overridden routes.

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

4. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_communities/communities/services/config.py
+++ b/invenio_communities/communities/services/config.py
@@ -42,7 +42,7 @@ from invenio_communities.communities.services.results import (
 from ...permissions import CommunityPermissionPolicy, can_perform_action
 from ..schema import CommunityFeaturedSchema, CommunitySchema, TombstoneSchema
 from .components import DefaultCommunityComponents
-from .links import CommunityLink
+from .links import CommunityLink, CommunityLinks
 from .search_params import IncludeDeletedCommunitiesParam, StatusParam
 from .sort import CommunitiesSortParam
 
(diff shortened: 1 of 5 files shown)


fenekku (commented):
Hey @monotasker just wanted to mention that I saw this (now). I am in the process of working on standardizing/unifying url generation in InvenioRDM and hopefully can integrate some of this PR in it ... eventually :sweat: (like end of month/March is my hope). Don't know if this is helpful in the immediate, but saw this by chance and thought better to mention it.

fenekku:
@monotasker Sorry for the delay on this! 

So with the new link generation and EndpointLink functionality in v13, the initial problem of [`COMMUNITIES_ROUTES` links](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/config.py#L22) not being picked up by the Resource's `link_items` [is resolved](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/services/config.py#L123). 

To be clear `COMMUNITIES_ROUTES` only applied to UI routes and still only does. Only those are impacted. There hasn't been a way to modify API routes across InvenioRDM as far as I know (invenio-rdm-records doesn't ...), so we will keep it that way for now. That change would need wider discussion. 

Let me know if that sounds ok and I will close this PR.

Thanks for your patience!

PR #1273, opened 2025-02-07
Humanize byte size error for community image upload

Meowcenary:

### Description

Update logo size limit error to show humanized byte size error on community image upload. Add helper function to perform this conversion.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_communities/errors.py
+++ b/invenio_communities/errors.py
@@ -12,6 +12,8 @@ from math import ceil
 from flask_babel import ngettext
 from invenio_i18n import gettext as _
 
+from .utils import humanize_byte_size
+
 
 class CommunityError(Exception):
     """Base exception for community errors."""
@@ -44,10 +46,15 @@ class LogoSizeLimitError(CommunityError):
 
     def __init__(self, limit, file_size):
         """Initialise error."""
+        limit_value, limit_unit = humanize_byte_size(limit)
+        file_size_value, file_size_unit = humanize_byte_size(file_size)
         super().__init__(
             _(
-                "Logo size limit exceeded. Limit: {limit} bytes Given: {file_size} bytes".format(
-                    limit=ceil(limit), file_size=ceil(file_size)
+                "Logo size limit exceeded. Limit: {limit_value} {limit_unit} Given: {file_size_value} {file_size_unit}".format(
+                    limit_value=limit_value,
+                    limit_unit=limit_unit,
+                    file_size_value=file_size_value,
+                    file_size_unit=file_size_unit,
                 )
             )
         )
--- a/invenio_communities/utils.py
+++ b/invenio_communities/utils.py
@@ -8,6 +8,8 @@
 
 """Utilities."""
 
+from decimal import ROUND_HALF_UP, Decimal
+
 from flask import session
 from flask_principal import Identity
 from invenio_accounts.models import Role
@@ -109,3 +111,14 @@ def on_datastore_post_commit(sender, session):
             users = role.users.all()
             for user in users:
                 on_user_membership_change(Identity(user.id))
+
+
+def humanize_byte_size(size):
+    """Converts bytes to largest unit (e.g., KB, MB, GB)."""
+    BYTES_PER_UNIT = 1000
+    s = Decimal(size)
+    q = Decimal("0.00")
+    for unit in ["B", "KB", "MB", "GB", "TB", "PB"]:
+        if s < BYTES_PER_UNIT:
+            return s.quantize(q, rounding=ROUND_HALF_UP), unit
+        s /= BYTES_PER_UNIT
--- a/tests/communities/test_resources.py
+++ b/tests/communities/test_resources.py
@@ -689,6 +689,9 @@ def test_logo_max_content_length(
         data=BytesIO(logo_data),
     )
     assert res.status_code == 400
+    assert (
+        res.json["message"] == "Logo size limit exceeded. Limit: 1.00 MB Given: 4.00 MB"
+    )
 
     # Update logo with small content
     logo_data = b"logo"


run-tests.sh
```diff
@@ -54,7 +54,7 @@ fi
 
 python -m check_manifest
 python -m sphinx.cmd.build -qnNW docs docs/_build/html
-eval "$(docker-services-cli up --db ${DB:-postgresql} --search ${SEARCH:-opensearch} --mq ${MQ:-redis} --env)"
+eval "$(docker services up --db ${DB:-postgresql} --search ${SEARCH:-opensearch} --mq ${MQ:-redis} --env)"
```
Meowcenary, line 57: The script run-tests.sh has a command that I don’t think is available on OSX. I tried making some changes to see if I could find a match for OSX, but that’s going nowhere. I think the move might be to spin up containers specifically for this, but I’m not sure where to find the files that define the containers for these tests. The --env flag seems to imply there is a Dockerfile or some other type of file that is supposed to be around. I was thinking it might make sense to just write my own and fill it in with the container definitions I need, but that seems wrong.

fenekku: Installing test dependencies should provide you with the `docker-services-cli` CLI utility: `pipenv run pip install -e .[tests,opensearch2]` . This is an invenio utility, not a docker provided thing and should work on MacOS. So these sections shouldn't be modified.

fenekku (commented):
Devil is in the details. Write up a test for the humanizing function and with suggestions/investigation we should have something that can be submitted :+1:

run-tests.sh
Meowcenary: I'll give that another shot. I was pretty sure I had run setup tools correctly, but there's a good chance I did something wrong and then went this painful road of ignorance and confusion.

Meowcenary (commented):
Thanks for all the feedback @fenekku! I'll add what I can to my ongoing notes and hopefully avoid some of this trouble next time.

invenio_communities/communities/services/config.py
```diff
@@ -19,7 +19,9 @@
     SearchOptionsMixin,
 )
 from invenio_records_resources.services.files.links import FileLink
-from invenio_records_resources.services.records.config import RecordServiceConfig
+from invenio_records_resources.services.records.config import (
```
Meowcenary, line 22: I'm not sure why this wasn't causing the test suite to fail, but it triggered when I was running the tests locally. I can leave this out if it doesn't make sense for this PR.

invenio_communities/utils.py
```diff
@@ -109,3 +110,16 @@ def on_datastore_post_commit(sender, session):
...
+    """Converts bytes to largest unit (e.g., KB, MB, GB)."""
+    old_rounding = getcontext().rounding
+    getcontext().rounding = ROUND_HALF_UP
+    s = Decimal(size)
+    for unit in ["B", "KB", "MB", "GB", "TB", "PB"]:
+        if s < 1000.0:
+            s = round(s, 2)
+            getcontext().rounding = old_rounding
+            return s, unit
+        s /= 1000.0
```
Meowcenary, line 125: I did not include a note about using a different library as a comment because I didn't want to overstep as a first time contributor. I get the feeling that regular contributors who have more experience will have opinions on this, so it might make more senes to discuss this in the PR.

Meowcenary (commented):
Updated and (finally) passing all tests locally. I was wondering where I should put a test for the humanize_byte_size function or if it was even something worth including?

I searched for some other functions in `utils.py` across the `tests` directory and was able to match a few to tests, but it doesn't seem like everything is unit tested, so I'm not sure if my changes are considered test worthy.

invenio_communities/communities/services/config.py
fenekku: Not sure why this was triggered locally for you: this line is 81 chars long, which is considered longer than the classic 80 of some python linting tools, but it is *not* too long for the `black` linter which we use. If you share what you ran and what you got, it may help. 

In any case, let's leave this line as it was before.

invenio_communities/errors.py
```diff
@@ -44,10 +46,15 @@ class LogoSizeLimitError(CommunityError):
 
     def __init__(self, limit, file_size):
         """Initialise error."""
+        limit_value, limit_unit = humanize_byte_size()
+        file_size_value, file_size_unit = humanize_byte_size()
         super().__init__(
             _(
-                "Logo size limit exceeded. Limit: {limit} bytes Given: {file_size} bytes".format(
-                    limit=ceil(limit), file_size=ceil(file_size)
+                "Logo size limit exceeded. Limit: {limit_value} {limit_unit} Given: {file_size_value} {file_size_unit} bytes".format(
```
fenekku, line 53:
```suggestion
                "Logo size limit exceeded. Limit: {limit_value} {limit_unit} Given: {file_size_value} {file_size_unit}".format(
```

invenio_communities/communities/services/config.py
utnapischtim: i think isort could caused the problem with this line like [here](https://github.com/inveniosoftware/invenio-communities/actions/runs/12999192145/job/36253989095?pr=1272#step:7:61693)

fenekku: We could try adding `line_length = 88` to the isort setup.cfg section:

```ini
# setup.cfg
[isort]
profile=black
line_length = 88
```

I guess profile=black isn't quite enough anymore?

invenio_communities/utils.py
```diff
@@ -109,3 +112,15 @@ def on_datastore_post_commit(sender, session):
             users = role.users.all()
             for user in users:
                 on_user_membership_change(Identity(user.id))
+
+
+def humanize_byte_size(size):
+    """Converts bytes to largest unit (e.g., KB, MB, GB)."""
+    s = Decimal(size)
+    for unit in ["B", "KB", "MB", "GB", "TB", "PB"]:
```
Meowcenary, line 120: It's probably unrealistic that anyone is going to upload anything that is even a gigabyte, let alone a terabyte or petabyte, but I don't think it adds a lot of overhead or anything, so it's likely fine.

invenio_communities/utils.py
```diff
@@ -109,3 +112,15 @@ def on_datastore_post_commit(sender, session):
...
+        if s < BYTES_PER_UNIT:
+            with localcontext() as ctx:
+                ctx.rounding = ROUND_HALF_UP
+                s = s.quantize(Decimal("0.00"))
```
Meowcenary, line 124: Does it make sense to have all calls to Decimal() use a string argument? It seems to be the best practice is to use a string, but from my perspective it doesn't make a difference for whole number values like 1000. I guess consistency would be a reason.

invenio_communities/utils.py
```diff
@@ -16,6 +18,7 @@
 from .generators import CommunityRoleNeed
 from .proxies import current_communities, current_identities_cache
 
+BYTES_PER_UNIT = Decimal(1000)
```
Meowcenary, line 21: This might be overkill or be better if it was renamed.

tests/communities/test_resources.py
```diff
@@ -689,6 +689,9 @@ def test_logo_max_content_length(
         data=BytesIO(logo_data),
     )
     assert res.status_code == 400
+    assert (
+        res.json["message"] == "Logo size limit exceeded. Limit: 1.00 MB Given: 4.00 MB"
```
Meowcenary, line 693: I'm going to look for other places that I could test this new function, but this seemed like a good spot to start. I'm not sure how thorough the tests needs to be.

Meowcenary (commented):
Hopefully close to being ready for official review.

invenio_communities/communities/services/config.py
utnapischtim: btw it is already fixed on the master branch. 
the `line_length = 88` seems not to work

invenio_communities/utils.py
```diff
@@ -109,3 +112,15 @@ def on_datastore_post_commit(sender, session):
...
+def humanize_byte_size(size):
+    """Converts bytes to largest unit (e.g., KB, MB, GB)."""
+    s = Decimal(size)
+    for unit in ["B", "KB", "MB", "GB", "TB", "PB"]:
+        if s < BYTES_PER_UNIT:
+            with localcontext() as ctx:
+                ctx.rounding = ROUND_HALF_UP
+                s = s.quantize(Decimal("0.00"))
+            return s, unit
+        s /= BYTES_PER_UNIT
```
fenekku, line 126: A couple last things:

- If we are going the `quantize` route, we may as well pass the rounding to simplify things: `s = s.quantize(q, rounding=ROUND_HALF_UP)` (`q = Decimal("0.00")` once and outside the loop) and remove all the context stuff
- Let's keep `BYTES_PER_UNIT` in the function. Was there a particular reason why out? (keeping it out exposes us to problems that we shouldn't have to care about)

Meowcenary: - Good call on the first point
- Honestly, the only reason I put it outside of the function was that I thought it was the standard way to place constants. I'm fine with scoping it to a function. For my own understanding, are we doing this to avoid naming collisions or were you thinking of something else?

invenio_communities/utils.py
fenekku: For small d decimal, a string is the way to go, but for whole numbers it doesn't matter too much like you are saying. It's not one of those consistency things that bothers me beyond that.

invenio_communities/communities/services/config.py
Meowcenary: Ah, thanks for the followup @utnapischtim

tmorrell (approved):
Thanks for these improvements! In the future I'd use more descriptive variables instead of s and q, but it's not worth going through another round of review.

PR #1291, opened 2025-04-04
using translated string for displaying role

Ducica:

### Description

There is an issue here that role is hardcoded in the members result list item https://github.com/inveniosoftware/invenio-communities/blob/7b9a1edd0e8a194e73349382d5e8320671c0bf90/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js#L142
image

Therefore, it is not possible to really change the content + not possible to localize.

This PR introduces a fix where we just additionally pass all roles to the ResultListItem and then we find the appropriate role's title. The users can configure the content of the text from invenio.cfg by using COMMUNITIES_ROLES variable.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js
@@ -148,6 +148,7 @@ export class ManagerMembersResultItem extends Component {
               label={i18next.t("Role") + " " + result.role}
             />
           ) : (
+            config?.allRoles?.find((role) => role.name === result.role)?.title ||
             _upperFirst(result.role)
           )}
         </Table.Cell>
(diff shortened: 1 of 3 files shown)


tmorrell (approved):
Thanks! Seems sensible to me, but it will need another review to get merged.

Ducica:
> Thanks! Seems sensible to me, but it will need another review to get merged.

Thanks Tom

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerMembersResultItem.js
```diff
@@ -148,6 +148,7 @@ export class ManagerMembersResultItem extends Component {
               label={i18next.t("Role") + " " + result.role}
             />
           ) : (
+            config?.allRoles?.find((role) => role.name === result.role)?.title ||
```
utnapischtim, line 151: could we move `config?.allRoles?.find((role) => role.name === result.role)?.title || _upperFirst(result.role)` before of the return statement. it is hard to read when this things are calculated directly within the `jsx` part.

Ducica: @utnapischtim Thanks, I made the requested change + I wrapped some strings used in UI in i18next. Hope that is OK. Please let me know.

Ducica:
@tmorrell Thank you very much!

PR #1292, opened 2025-04-04
permissions: replace Disabled with SystemProcess

ntarocco:
* the Disabled permission does not allow admins to perform operations


ptamarit:
Put back to in progress since there are some broken tests.

PR #1294, opened 2025-04-25
dependencies: depend explicitly on invenio-records-resources

fenekku:


tmorrell (approved):
This would have saved a lot of trouble with the axios PR

utnapischtim (approved):

PR #1299, opened 2025-05-04
fix: rename variable

utnapischtim:
* the variable name in communities_requests is community and not
  community_ui

### DIFF
--- a/invenio_communities/views/communities.py
+++ b/invenio_communities/views/communities.py
@@ -2,7 +2,7 @@
 #
 # This file is part of Invenio.
 # Copyright (C) 2016-2024 CERN.
-# Copyright (C) 2023 Graz University of Technology.
+# Copyright (C) 2023-2025 Graz University of Technology.
 # Copyright (C) 2024 KTH Royal Institute of Technology.
 #
 # Invenio is free software; you can redistribute it and/or modify it
@@ -309,7 +309,6 @@ def communities_new_subcommunity(pid_value, community, community_ui):
     return render_community_theme_template(
         "invenio_communities/details/new_subcommunity.html",
         theme=community_ui.get("theme", {}),
-        community=community,
         community_ui=community_ui,
         permissions=permissions,  # hide/show UI components
         form_config=dict(
@@ -334,7 +333,7 @@ def communities_subcommunities(pid_value, community, community_ui):
     return render_community_theme_template(
         "invenio_communities/details/subcommunity/index.html",
         theme=community_ui.get("theme", {}),
-        community=community_ui,
+        community_ui=community_ui,
         permissions=permissions,
     )
 
@@ -370,7 +369,6 @@ def communities_settings(pid_value, community, community_ui):
     return render_community_theme_template(
         "invenio_communities/details/settings/profile.html",
         theme=community_ui.get("theme", {}),
-        community=community,
         community_ui=community_ui,
         has_logo=True if logo else False,
         logo_quota=logo_size_limit,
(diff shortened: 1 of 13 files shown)


invenio_communities/views/communities.py
```diff
@@ -392,6 +392,7 @@ def communities_requests(pid_value, community, community_ui):
         "invenio_communities/details/requests/index.html",
         theme=community_ui.get("theme", {}),
         community=community_ui,
+        community_ui=community_ui,
```
ntarocco, line 395: Having the 2 vars with the same value is not super great. They have different purposes, see for example [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/views/communities.py#L312).
I would suggest changing instead:
- the current var in the previous line, which is clearly wrongly named:
```diff
- community=community_ui,
+ community_ui=community_ui,
```
- its usage in the template [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/requests/index.html#L27)

WDYT?

utnapischtim: the reason why i did the quick fix and add another variable with the same name and value was because i couldn't understand where `community` is used. since the refactor commit https://github.com/inveniosoftware/invenio-communities/commit/9909924edd15abfde7137a4a9b67b3e8e6c9da22 introduced that bug, i thought it is safer to not remove `community=community_ui`

utnapischtim: so to be safe, all `community` variables would have to be renamed to `community_ui` in the context of `invenio_communities/views/communities.py` and the then used jinja templates?

ntarocco: > the reason why i did the quick fix and add another variable with the same name and value was because i couldn't understand where `community` is used. since the refactor commit [9909924](https://github.com/inveniosoftware/invenio-communities/commit/9909924edd15abfde7137a4a9b67b3e8e6c9da22) introduced that bug, i thought it is safer to not remove `community=community_ui`

It should be in the view's template (I linked it above).

> so to be safe, all `community` variables would have to be renamed to `community_ui` in the context of `invenio_communities/views/communities.py` and the then used jinja templates?

Only the ones that have value `community_ui` -> `community=community_ui`. Basically, community is the obj, and community_ui it serialized version, they are different. We should keep naming consistent to mark the difference.

utnapischtim: > Only the ones that have value `community_ui` -> `community=community_ui`. Basically, community is the obj, and community_ui it serialized version, they are different. We should keep naming consistent to mark the difference.

i think that doesn't work either without putting more thinking and testing time into it. for [communities_subcommunities](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/views/communities.py#L337) `community` is used [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html#L14) and [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/subcommunity/index.html#L37). can i replace that with `community_ui["metadata"]["title"]` and `community_ui["slug"]`?

btw this is also a problem [here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/templates/semantic-ui/invenio_communities/details/settings/base.html#L11)

so it could be that i need community=community and community_ui=community_ui, or i refactor the usage of community.metadata etc completly

ntarocco: For the time being, fine by me adding both:
```python
community=community
community_ui=community_ui
```
☺️

PR #1307, opened 2025-06-12
fixing missing translations in UI

Ducica:
:heart: Thank you for your contribution!

### Description

> Please describe briefly your pull request.

We are trying to correct parts of the UI that are not properly localized. This PR deals with:

image

image

On communities settings page, the organization selector:
image

I have tried to keep the changes to a minimum. After testing, from my point of view, it seems nothing else broke.

NOTE: I have decided to remove the trans component because in my opinion, it cannot work like this i.e. there is no i18next instance provided to it + the search app is not wrapped in I18nextProvider and I think it just adds complexity and does not really solve anything.

Please let me know what you think and thanks.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
[...]

PR #1310, opened 2025-06-16
translations: i18next.t instead of Trans for resultsPerPage component

Ducica:

### Description

NOTE: this was derived from https://github.com/inveniosoftware/invenio-communities/pull/1307
where based on comments from Tom and Christoph I will split that PR into more self contained PRs for easier reviewing and merging. After I make all derived PRs, I will close the original one.

> Please describe briefly your pull request.

Results per page component is not localized properly (members and invitations pages):

image

image

The reason is that Trans component is used incorrectly:

```
  <Trans key="communitiesMembersResult" count={cmp}>
                  {cmp} results per page
                </Trans>
```

1.  Key is incorrect prop - should be i18nKey
2.  You cannot pass a component as a prop, but rather some interpolated value 
3. These two search apps are not wrapped by I18nextProvider and this means, if you wish for ```<Trans/>```
to work properly, you must pass it i18n prop (specific i18next instance).

From my point of view, I see no added benefits for using Trans here, just added complexity, so my proposal is to just use i18next.t instead.

**Reminder**

5. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationsResults.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/InvitationsResults.js
@@ -10,7 +10,7 @@ import React from "react";
 import { Grid } from "semantic-ui-react";
 import { ResultsPerPage, Pagination, ResultsList } from "react-searchkit";
 import PropTypes from "prop-types";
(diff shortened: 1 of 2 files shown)


utnapischtim:
superseded by #1313

PR #1312, opened 2025-06-24
feat (services): allow custom schema and record class

egabancho:
What would be the best way to document these configuration variables?

Inside RDM-Records, there is no reference to their counterpart in `config.py`, yet I feel it might be a good idea to add them there with the default value, so that they are somehow discoverable.

### DIFF
--- a/invenio_communities/communities/services/config.py
+++ b/invenio_communities/communities/services/config.py
@@ -90,7 +90,7 @@ class CommunityServiceConfig(RecordServiceConfig, ConfiguratorMixin):
         "COMMUNITIES_PERMISSION_POLICY", default=CommunityPermissionPolicy
     )
     # Record specific configuration
-    record_cls = Community
+    record_cls = FromConfig("COMMUNITIES_COMMUNITY_CLS", default=Community)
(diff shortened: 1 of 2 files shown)


invenio_communities/communities/services/config.py
```diff
@@ -95,7 +95,7 @@ class CommunityServiceConfig(RecordServiceConfig, ConfiguratorMixin):
         "COMMUNITIES_PERMISSION_POLICY", default=CommunityPermissionPolicy
     )
     # Record specific configuration
-    record_cls = Community
+    record_cls = FromConfig("COMMUNITIES_COMMUNITY_CLS", default=Community)
```
kpsherva, line 93: looks OK to me, however this configuration variable does not exist

egabancho: That's exactly what I asked in my initial comment ☺️ . The [RDM-Records](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L396) equivalent doesn't exist anywhere else either.

PR #1350, opened 2026-01-30
feat: added overridable blocks around community invitations modal

Ducica:

### Description

It is not possible currently to override the community invitations modal. This PR adds overridable block around each instance where the community invitations modal appears.

For example, in our case, the community addition is done in an external system and then sychronized back to the repository and it is beneficial that we are able to provide our own UI for the invitations modal.

Thanks a lot for taking a look.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerSearchLayout.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerSearchLayout.js
@@ -15,6 +15,7 @@ import { SearchFilters } from "@js/invenio_search_ui/components";
 import { SearchBar, Sort } from "react-searchkit";
 import { InvitationsContextProvider } from "../../../api/invitations/InvitationsContextProvider";
 import { InvitationsMembersModalWithSearchKit } from "../../invitations/invitationsModal/InvitationsMembersModal";
+import Overridable from "react-overridable";
 
 export class ManagerSearchLayout extends Component {
   render() {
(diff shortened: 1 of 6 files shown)


tmorrell (commented):
It's generally good to add more overridable labels, but shouldn't they be specific to the component? https://inveniordm.docs.cern.ch/community/code/best-practices/react/#overridable-components-naming @kpsherva can provide more guidance that I'm able to on the correct naming scheme.

invenio_communities/assets/semantic-ui/js/invenio_communities/members/members/manager_view/ManagerSearchLayout.js
```diff
@@ -29,11 +30,18 @@ export class ManagerSearchLayout extends Component {
           <div>
             <div className="mobile only rel-mb-1">
               <InvitationsContextProvider community={community}>
-                <InvitationsMembersModalWithSearchKit
+                <Overridable
+                  id="InvenioCommunities.CommunityMembersSearch.ManagerSearchLayout.InvitationsModal"
```
kpsherva:
```suggestion
                  id="InvenioCommunities.CommunityMembersSearch.ManagerSearchLayout.InvitationsModal.container"
```

if I remember correctly (please verify with the documentation), when we add an overridable wrapper of a component, then we append container to the ID (the same comment applies to other occurences in this PR)

Ducica: Thanks. I should have been more careful. Just a note, https://inveniordm.docs.cern.ch/community/code/best-practices/react/#overridable-components-naming I am not sure if this is the documentation, but here it is a bit of a mess Container vs container. Otherwise, I checked the codebase and it is .container. Again thanks for the time.

kpsherva (approved):

PR #1351, opened 2026-02-02
feat: added blocks to communities templates

Ducica:

### Description

We make a lot of apps based on Invenio system, and sometimes. We try to maximise reuse of existing front and. It is sometimes necessary to override certain behaviors, but then we need to copy paste the entire template and modify, which is not ideal and we also later lose access to new features as templates evolve. We think that adding overridable blocks to these templates is also generally beneficial for the ecosystem, but for us we have concrete needs to override these blocks.

Please if possible to merge this.

Thank you,

Dusan S (CESNET)


tmorrell (approved):
Looks good to me. @kpsherva might have thoughts on naming.

utnapischtim (approved):

PR #1361, opened 2026-03-20
feat(manager): Add API to search accepted community requests

### LINKED ISSUE #568: "View comments" button under the community is not visible on other record versions (landing page)
## Is your feature request related to a problem? Please describe.
Currently, if a request is accepted for a record version, the "View Comments" button under the accepted community only shows up for that specific version. It is stored in the "topic" sub-field of the request's payload.

## Describe the solution you'd like
There should be a way to view the request comments from the landing page of all versions.

This will require to have the parent_id on the requests index, to be able to show the button on all other versions.

### Solution proposal (to be discussed and finalized)
The link can redirect to the dashboard with the results containing all the related requests on the parent id.
This way if there is scenario where a community inclusion request of version was not accepted but the version 2 was accepted, the link will redirect to the requests dashboard containing all requests for those records in that community to be able to refer to the older request for reference.

#### Steps
1. Add parent id to the request index and update index (all new requests will contain parent_id)
2. Re-index the requests (where parent_id doesn't exist, all requests before we updated the index)
3. Update the link in the UI at the end (this way we avoid any searches on the parent till we have it indexed on all)

## Additional context
Another thing to consider is that on the request dashboard currently there is no way to view all open/closed/owned/shared request in a single list view. We can have a way to be able to view.

sakshamarora1:
fixes: https://github.com/inveniosoftware/invenio-requests/issues/568

### DIFF
--- a/invenio_communities/records/records/systemfields/communities/manager.py
+++ b/invenio_communities/records/records/systemfields/communities/manager.py
@@ -58,7 +58,7 @@ class CommunitiesRelationManager:
     # API
     #
     def add(self, community_or_id, request=None, default=False):
-        """Add a record to a community.
+        """Add a record (parent) to a community.
 
         If a record was already added to a community an IntegrityError will
         be raised.
@@ -85,7 +85,7 @@ class CommunitiesRelationManager:
             self._communities_cache[community_id] = community_or_id
 
     def remove(self, community_or_id):
-        """Remove a record from a community."""
+        """Remove a record (parent) from a community."""
         community_id = self._to_id(community_or_id)
 
         # Delete M2M row.
@@ -103,7 +103,7 @@ class CommunitiesRelationManager:
             self._default_id = None
 
     def clear(self):
-        """Clear all communities from the record."""
+        """Clear all communities from the record (parent)."""
         # Remove all associations
         res = self._m2m_model_cls.query.filter_by(record_id=self._record_id).delete()
         self._communities_ids = set()
@@ -125,6 +125,14 @@ class CommunitiesRelationManager:
         # Unset default if no longer available
         if self._default_id and self._default_id not in self._communities_ids:
             self._default_id = None
+
+    def get_requests(self):
+        """Get all accepted requests for the record (parent)."""
+        # Only requests that led to the record being added to the community are stored in the table
+        return self._m2m_model_cls.query.filter(
+            self._m2m_model_cls.record_id == self._record_id,
+            self._m2m_model_cls.request_id.isnot(None),
+        ).all()
 
     def __len__(self):
         """Get number of communities."""


kpsherva (approved):

invenio_communities/records/records/systemfields/communities/manager.py
```diff
@@ -126,6 +126,11 @@ def refresh(self):
         if self._default_id and self._default_id not in self._communities_ids:
             self._default_id = None
 
+    def get_accepted_requests(self):
+        """Get all accepted requests for the record (parent)."""
+        # Only requests that led to the record being added to the community are stored in the table
+        return self._m2m_model_cls.query.filter_by(record_id=self._record_id).all()
```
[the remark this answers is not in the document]
slint: I think we're still missing an index on `record_id`... Just for reference, this query on Zenodo prod **takes 3sec** for [this record](https://zenodo.org/records/20024933):

```sql
zenodo_read@dbod-zenodo-prod.cern.ch:zenodo> select * from rdm_parents_community where record_id = '3c78
 1ce3-5b7b-4e9a-ac42-c2aac5573f5e' and request_id is not null;
-[ RECORD 1 ]-------------------------
community_id | 00f1c90f-7394-48cc-b96d-c62071a78597
record_id    | 3c781ce3-5b7b-4e9a-ac42-c2aac5573f5e
request_id   | 7fbcf2ad-9fe4-4952-8525-f23f70a4a420
-[ RECORD 2 ]-------------------------
community_id | f0a8b890-f97a-4eb2-9eac-8b8a712d3a6c
record_id    | 3c781ce3-5b7b-4e9a-ac42-c2aac5573f5e
request_id   | 7fbcf2ad-9fe4-4952-8525-f23f70a4a420
SELECT 2
Time: 3.202s (3 seconds), executed in: 3.202s (3 seconds)
```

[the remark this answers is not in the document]
slint: @sakshamarora1 did you fix the column index in another PR? I think this is the last thing blocking this from being merged on my side. We also need an Alembic recipe (in invenio-rdm-records, where the actual models/tables are defined), to add the index.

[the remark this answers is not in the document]
sakshamarora1: @slint I have added the alembic recipe, here in the rdm-records PR: https://github.com/inveniosoftware/invenio-rdm-records/pull/2294/commits/a5462b75905992fd81cd0c7a0fc851c68be18505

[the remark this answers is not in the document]
slint: Thanks, I missed that one. We still need to have the index though [on the mixin class column in this module](https://github.com/inveniosoftware/invenio-communities/blob/0824825e4e67da7675689dad2f637f5cbd38fc36/invenio_communities/records/records/models.py#L47).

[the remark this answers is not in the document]
sakshamarora1: Ah yes thank you! Added now

invenio_communities/records/records/models.py
```diff
@@ -46,6 +46,7 @@ def record_id(cls):
             UUIDType,
             db.ForeignKey(cls.__record_model__.id, ondelete="CASCADE"),
             primary_key=True,
+            index=True,
```
kpsherva: don't we need an alembic recipe for this change?

tests/test_alembic.py
```diff
@@ -36,12 +36,12 @@ def test_alembic(base_app, database, extra_entry_points):
     ext.alembic.upgrade()
 
     # expected non-covered migrations (from previous tests that use mock_module):
-    # ('add_table', Table('mock_metadata'
-    # ('add_table', Table('mock_community'
+    expected_migrations = {
```
kpsherva: not sure why this modification is needed?

utnapischtim: the `add_index` part exists, because the recipe for it is in rdm-records. probably the recipe should be moved to this package?

invenio_communities/records/records/models.py
kpsherva: @mesemus @utnapischtim would it work for you if we move this to rdm-records? otherwise it is causing confusion and problems on the migration recipes as we see in my comment above :)

mesemus: The idea is that every class inheriting from `CommunityRelationMixin` would automatically get the index on `record_id`, which is a good thing, and the Alembic migration in the subclass' package would take care of creating the index. In RDM, there is currently only a single subclass: RDMParentCommunity.

To minimize confusion, I think it's okay to move it there - see my comment in the other PR: https://github.com/inveniosoftware/invenio-rdm-records/pull/2294#issuecomment-4631247835

(And yes, I was confused as well, this time from the other side.)

The disadvantage of this approach is the need to use a different form of index declaration due to the fact that the field is defined on the mixin. 

If we're not planning to add any other "objects" to communities, I'd vote for moving it. For the CESNET overlay, it's not a problem, as all our records that can belong to a community inherit from the RDM record.

tests/test_alembic.py
mesemus: This modification is needed because [MockRecordCommunity](https://github.com/inveniosoftware/invenio-communities/blob/master/tests/records/mock_module/models.py#L15) gets an index on `record_id` from this pull request, but no Alembic migration is generated for `MockRecordCommunity` as it is just a test mock orm table. As a result, if we start with an empty database and call `ext.alembic.upgrade()`, the index is not created. Consequently, `compare_metadata()`, which compares the database schema with the SQLAlchemy model definitions, reports a discrepancy.

This is why `ix_mock_community_record_id` was added to the ignore list.

mesemus: > probably the recipe should be moved to this package

we cannot do this as that would mean that alembic migrations in communities would become dependent on invenio-rdm-records orm classes. 

I guess that the safest way would be to move the index to rdm records as Karolina suggested above.

invenio_communities/records/records/models.py
mesemus: If we move the index, it would be nice to have a comment here in docstring - something like: in inherited subclasses please add index on record_id if you plan to look up record's communities.

slint: Just chipping in, since it was my idea behind the scenes to move the `CommunityRelationMixin` to `invenio-rdm-records`. In terms of executing this:

1. Add to `rdm-records` the mixin class and replace the import.
1. To be confirmed by @utnapischtim and @mesemus, :+1: this comment if you agree on this point: Can we just add `index=True` on the mixin, instead of adding `Index("ix_rdm_parents_community_record_id", RDMParentCommunity.record_id)` on `RDMParentCommunity`? I guess for you it means that if you have other `RDMParentCommunity`-like models on your instances, you would need to create an Alembic recipe for each.
1. (Optional) Deprecate `CommunityRelationMixin` and release a minor
1. Remove `CommunityRelationMixin` entirely from this module.

The point of all the above is that we can get away with a minor release on `invenio-rdm-records`, given that we're just talking about a column index (i.e. a performance fix) and not any functional/relational change to the schema.

utnapischtim: i see no option to release this whole change as a minor in rdm-records. there is at least one alembic recipe involved and therefore a major bump is the way to go.

> I guess for you it means that if you have other RDMParentCommunity-like models on your instances, you would need to create an Alembic recipe for each.

as i see the whole thing we would have to do that anyway. the alembic recipe is in rdm-records because the model is there, and therefore if we would have a model like `RDMParentComunity` we would need to write our own alembic recipe anyway.

in my point of view the problem is in the to much coupling of rdm-records with communities, but we can't solve that part now.

in my opinion moving `CommunityRelationMixin` to rdm-records is a good idea. so we have everything in one place, the index=True addition and the alembic recipe

mesemus: @slint @utnapischtim Agree, I have no problem moving the CommunityRelationMixin to invenio-rdm-records, and then index=True on it is completely ok - both the index and the alembic will be within a single library (and a single PR).

I would go for 4 - the mixin was used only in rdm-records, so imho it would be ok not to deprecate it, have a major version here in communities with the mixin removed (as it looks we will need to have a major release in rdm-records, we could do it as well here ?).

sakshamarora1: I have added a new PR with the final decision implemented: https://github.com/inveniosoftware/invenio-rdm-records/pull/2361
Here is the communities PR for deprecating the `CommunityRelationMixin` class https://github.com/inveniosoftware/invenio-communities/pull/1390
I have also removed the Alembic recipe and the `index=True` out of these 2 PRs to the new rdm-records PR

Now all of these PRs can be merged independently.

tests/test_alembic.py
sakshamarora1: Please see my reply with the updated state: https://github.com/inveniosoftware/invenio-communities/pull/1361#discussion_r3396170345

PR #1369, opened 2026-04-10
fix: use baserequest for subcommunity requests

OliverGeneser:


invenio_communities/subcommunities/services/request.py
```diff
@@ -9,7 +9,7 @@
 from invenio_access.permissions import system_identity
 from invenio_i18n import lazy_gettext as _
 from invenio_notifications.services.uow import NotificationOp
-from invenio_requests.customizations import RequestType, actions
+from invenio_requests.customizations import BaseRequest, actions
```
slint: This is in `rdm-records`... which of course is not a dependency of `invenio-communities`

OliverGeneser: Just saw that...

slint (approved):
LGTM!

PR #1376, opened 2026-04-22
fix(inv): show discussion for invitation requests by generating self_html

fenekku:
Fixes issue with invitation discussion page not rendering comments or request sidebar.

### DIFF
--- a/invenio_communities/members/services/request.py
+++ b/invenio_communities/members/services/request.py
@@ -22,6 +22,7 @@ from invenio_communities.notifications.builders import (
     CommunityInvitationExpireNotificationBuilder,
 )
 
+from ...communities.services.service import get_cached_community_slug
 from ...proxies import current_communities
 
 
@@ -102,6 +103,60 @@ class ExpireAction(actions.ExpireAction):
 #
 # Request
 #
+def is_created_by_current_user(obj, vars):
+    """Is created by current user.
+
+    :param request: api.Request
+    :param vars: dict: context variables w/ keys:
+        "permission_policy_cls" for api.Request
+        "identity"
+        "request"
+        "request_type"
+    """
+    request = vars["request"]
+    id_creator = str(request.created_by.reference_dict.get("user"))
+    id_identity = str(vars["identity"].id)
+    return id_identity == id_creator
+
+
+def update_vars_for_user_dashboard_request_view(obj, vars):
+    """Update vars.
+
+    Because this is called in the context of a RequestType, the received obj
+    can be a Request or RequestEvent so better to trust the vars content.
+
+    :param obj: api.Request|api.RequestEvent
+    :param vars: dict: context variables w/ keys:
+        "permission_policy_cls" for api.Request
+        "identity"
+        "request"
+        "request_type"
+    """
+    vars.update(request_pid_value=str(vars["request"].id))
+
+
+def update_vars_for_community_dashboard_request_view(obj, vars):
+    """Update vars.
+
+    Because this is called in the context of a RequestType, the received obj
+    can be a Request or RequestEvent so better to trust the vars content.
+
+    :param obj: api.Request|api.RequestEvent
+    :param vars: dict: context variables w/ keys:
+        "permission_policy_cls" for api.Request
+        "identity"
+        "request"
+        "request_type"
+    """
+    request = vars["request"]
+    # The topic holds a CommunityPKProxy
+    slug = get_cached_community_slug(request.topic.reference_dict["community"])
+    vars.update(
+        pid_value=slug,
+        request_pid_value=request.id,
+    )
+
+
 class CommunityInvitation(RequestType):
     """Community member invitation request type."""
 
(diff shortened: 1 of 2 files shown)


invenio_communities/members/services/request.py
```diff
@@ -102,6 +103,60 @@ def execute(self, identity, uow):
...
+    :param obj: api.Request|api.RequestEvent
+    :param vars: dict: context variables w/ keys:
+        "permission_policy_cls" for api.Request
+        "identity"
+        "request"
+        "request_type"
+    """
+    request = vars["request"]
+    # The topic holds a CommunityPKProxy
+    slug = get_cached_community_slug(request.topic.reference_dict["community"])
```
fenekku, line 153: This is the "best" we can do given that the community slug is not available in this context and we need it for a UI URL. It will make at most 1 DB call (and none if already cached).

invenio_communities/members/services/request.py
```diff
@@ -131,6 +186,23 @@ class CommunityInvitation(RequestType):
...
+        "self_html": ConditionalLink(
+            cond=is_created_by_current_user,
+            if_=EndpointLink(
+                "invenio_app_rdm_requests.user_dashboard_request_view",
+                params=["request_pid_value"],
+                vars=update_vars_for_user_dashboard_request_view,
+            ),
+            else_=EndpointLink(
+                # Ideally have a _WithFallbackEndpointLink()
+                "invenio_app_rdm_requests.community_dashboard_request_view",
```
fenekku, line 199: The `/communities/<community pid>/invitations/<pid>` route is currently served by the same view as the one serving `/communities/<community pid>/requests/<pid>` in invenio-app-rdm: https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/requests_ui/views/ui.py#L50-L57 . This makes generating the `/invitations` route impossible currently, but the `/communities/<community_pid>/requests/<pid>` generated route will render the discussion page correctly.

In a future PR in invenio-app-rdm I will extract the invitation related code from  `community_dashboard_request_view` and introduce a "fallback EndpointLink" here to make the transition seamless.

invenio_communities/members/services/request.py
slint: Isn't it that community UUIDs also work with URLs? I see that [the `PIDSlugField` handles both UUIDs and slugs](https://github.com/inveniosoftware/invenio-communities/blob/bc8291ec9b38b8daf5c36c11a91bbeaee3020f80/invenio_communities/communities/records/systemfields/pidslug.py#L35-L41). In that case maybe changing the view logic to be more lenient is better?

Ideally, I would prefer if the UUID URLs resolved + redirected to the slug version, but that's a bigger change.

invenio_communities/members/services/request.py
slint: But isn't it that we can provide a unique name/ID when registering each endpoint, and thus allow being specific about which path we prefer in `(invenio_)url_for`? If that's possible, I would then keep `community_dashboard_request_view` for the `.../requests/...` variant (for backwards compat), an make a new endpoint name/ID for the invitations.

slint (approved):
LGTM. Maybe one thing to check is that in email notifications we sometimes hardocde the desired request UI path URL, instead of using `request.links.self_html` (or similar). This has been in general problematic because we don't serialize the links based on the identity of the notification recipient...

invenio_communities/members/services/request.py
fenekku: > Isn't it that community UUIDs also work with URLs?

Yes they do. Here we are trying to make sure that any UI link uses the community slug rather than the UUID for UI/UX/branding/sharing purposes. It's a bad UX when the browser bar shows `/communities/my-community/...` and shifts to `/communities/12345.../...` when clicking on some links. Given the vanity URL is important enough to have a form field at community creation, it felt like it had to be respected.

> Ideally, I would prefer if the UUID URLs resolved + redirected to the slug version, but that's a bigger change.

Yes! That's the ideal change and I got something similar coming for the case of redirecting `/communities/.../requests/...` to `/communities/.../invitations/...` in the case of an invitation request.

invenio_communities/members/services/request.py
fenekku: Yes [`add_url_rule`](https://flask.palletsprojects.com/en/stable/api/#flask.Flask.add_url_rule) does allow passing an `endpoint`, but invenio-communities can't count on that new endpoint name since the UI endpoint is defined in invenio-app-rdm *which depends on* invenio-communities. So in the PR I have in the wings for invenio-app-rdm, as mentioned above, I make `/commnities/.../requests/...` in the case of an invitation request redirect to `/communities/.../invitations/...` (great minds think alike, I already had the idea —so we're clear hahaha :joy_cat: !— and implemented it, it's just not out yet...).

The FallbackEndpointLink approach proved to be needlessly complex compared to that. Comment will be cleaned up in the PRs I have already lined up next to be sent.

fenekku:
> LGTM. Maybe one thing to check is that in email notifications we sometimes hardocde the desired request UI path URL, instead of using `request.links.self_html` (or similar). This has been in general problematic because we don't serialize the links based on the identity of the notification recipient...

I am actually wrapping up notifications for membership requests right now and I have been looking at the invitations notifications and record submission ones. By design, so far, it seems like those templates (especially invitations I know) are written and sent with a specific recipient targeted. For example, the "community-invitation.accept.jinja" template is only sent to the community managers, so the link there can be explicitly for the community dashboard and so on. I haven't encountered yet a scenario where the same email is sent to parties on either side of a request. If the case happens, I will reach out on discord... I am wandering about best/better architecture of these cases recently...

PR #1381, opened 2026-05-07
fix(community-themes): only request theme CSS for themed communities

max-moser:
Previously, the community base template *always* requested the community theme CSS, regardless of the community actually being themed or not.
For un-themed communities, this resulted in unnecessary requests with zero-length responses.

This PR requests community theme CSS only if the community has a theme set.


Samk13 (approved):

fenekku (approved):
🙏 we kept putting off fixing this on our side for other tickets. Thanks for taking care of it!

PR #1395, opened 2026-06-23
Removing `partial()` calls in marshmallow schemas

max-moser:
Based on @sakshamarora1 's PR https://github.com/inveniosoftware/invenio-communities/pull/1388 but without the schema registry

### DIFF
--- a/invenio_communities/communities/resources/ui_schema.py
+++ b/invenio_communities/communities/resources/ui_schema.py
@@ -1,11 +1,9 @@
-# SPDX-FileCopyrightText: 2022-2024 CERN.
-# SPDX-FileCopyrightText: 2023 TU Wien.
+# SPDX-FileCopyrightText: 2022-2026 CERN.
+# SPDX-FileCopyrightText: 2023-2026 TU Wien.
 # SPDX-FileCopyrightText: 2025 Graz University of Technology.
 # SPDX-License-Identifier: MIT
 
 """UI community schema."""
-
-from functools import partial
 
 from flask import g
 from flask_resources import BaseObjectSchema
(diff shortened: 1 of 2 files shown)


invenio_communities/communities/resources/ui_schema.py
```diff
@@ -57,10 +55,6 @@ def mask_removed_by(obj):
     return return_value
 
 
-# Partial to make short definitions in below schema.
-FormatEDTF = partial(FormatEDTF_, locale=get_locale)
```
utnapischtim, line 61: question: is this necessary to make it possible to use the GlobalSchemaRegistry?

sakshamarora1: Yes, removing the partial calls everywhere ensures that the schema holds a static reference to the nested schema instance which is what these changes aim for to do away with the request time schema instantiation (and later on for DeepFriendMarshmallow if/when added)

P.S. Also the GlobalSchemaRegistry was removed with Max's updates in this new PR, but the global singleton design is still there by keeping the schema instantiated and saved on the object

utnapischtim: ok, this means, the constructor of FormatEDTF is now called once? probably on app creation?  before more often? is there a reason why this hasn't been done before?

max-moser: note: the previous `FormatEDTF` definition here was just a convenience utility for later usage down below (saved one argument)

the `partial()` calls got removed as preparation for DFM (what @sakshamarora1 said) and because they were generally not required and added a layer of untouchable mystery
i couldn't find a good reason for it being done that way in the first place; it seems like that was introduced at some point somewhere and then just copied all over the place

regarding one-time init:
the schemas are now constructed only once, in the services' constructors; as opposed to a new schema on each service call before
the reason why it was created so often before was probably the dynamic context, which got replaced by the context var

ciurana-life (approved):

OliverGeneser (approved):
lgtm

PR #1398, opened 2026-07-06
fix(members): debounce user/group search in invitation and share modals

slint:
* Search fired one request per keystroke; now debounced (500ms).
* Cancel in-flight requests so stale responses don't overwrite newer
  results, and clean up on unmount.

### DIFF
--- a/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js
+++ b/invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js
@@ -8,6 +8,7 @@ import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { Image, withCancel } from "react-invenio-forms";
 import { Dropdown, Grid, Header } from "semantic-ui-react";
+import _debounce from "lodash/debounce";
 import _truncate from "lodash/truncate";
 
 export class MembersSearchBar extends Component {
@@ -138,12 +139,20 @@ export class MembersSearchBar extends Component {
     handleChange(selectedMembers);
   };
 
-  onSearchChange = async (event, { searchQuery }) => {
+  componentWillUnmount() {
+    this.onSearchChange.cancel();
+    this.cancellableSuggestions?.cancel();
+  }
+
+  onSearchChange = _debounce(async (event, { searchQuery }) => {
     const { fetchMembers } = this.props;
+    // cancel in-flight requests so stale responses don't overwrite newer ones
+    this.cancellableSuggestions?.cancel();
+    const cancellableSuggestions = withCancel(fetchMembers(searchQuery));
+    this.cancellableSuggestions = cancellableSuggestions;
     try {
       this.setState({ isFetching: true });
 
-      const cancellableSuggestions = withCancel(fetchMembers(searchQuery));
       const suggestions = await cancellableSuggestions.promise;
       this.setState({
         isFetching: false,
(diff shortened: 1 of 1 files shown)


invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js
```diff
@@ -138,26 +139,34 @@ export class MembersSearchBar extends Component {
...
         error: false,
       });
     } catch (e) {
+      if (e === "UNMOUNTED") return;
       console.error(e);
       this.setState({
         isFetching: false,
         error: true,
       });
     }
```
OliverGeneser, line 174: nit; we could clean up the state a bit by restting the reference after the promise settles. 
```finally { this.cancelableSuggestions = null; }```

File: invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js [hunk 2]
```diff
@@ -138,26 +139,34 @@ export class MembersSearchBar extends Component {
...
-  };
+  }, 500);
```
OliverGeneser: Does 500ms feel good? I would think it might be a bit too high and make the UI feel slightly sluggish 🤔

OliverGeneser (approved):

invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js
OliverGeneser: Personally I would decrease it to 250-300ms

slint: Apparently, this is the [default on RemoteSelectField](https://github.com/inveniosoftware/react-invenio-forms/blob/82d57a9174d016f23395e3896912f6a7c98d684c/src/lib/forms/RemoteSelectField.js#L336C3-L336C15) (and is not overridden anywhere from a quick search).

But agree that 300ms is much better, and lower than [the 400ms which is based on some research on UX responsiveness](https://lawsofux.com/doherty-threshold/).

invenio_communities/assets/semantic-ui/js/invenio_communities/members/invitations/invitationsModal/MemberSearchBar.js
slint: Clanker suggested that just adding in the finally is subject to a race condition edge-case, so there are some whacky captured values now.

I can take it for a spin, but much later in the day...

OliverGeneser: Ohh please let me know when/if you find the race condition edge-case that clanker mentions

slint: Clanker's words (I'll try to reproduce by setting 3G network throttle):

> when a new search cancels an in-flight one, the cancelled request's `finally` runs after the new request has stored its reference, and would null it out, silently disabling cancellation for the next keystroke. It needs a local capture + identity check (`if (this.cancellableSuggestions === cancellable) ...`). (FWIW the stale reference was harmless, `cancel()` on a settled promise is a no-op.)

OliverGeneser: Oh I see what your Clanker is saying... Next time I will ask Chatty first :sweat_smile:

